PR #1, opened 2024-01-30
Fix broken Hugging Face link

thenameless7741:
The [original link](https://huggingface.co/models/argmaxinc/whisperkit-coreml) to the models on Hugging Face website was broken, so this PR just fixes it.

Edit 1: I didn't have `git-lfs` installed when I ran `download-models`, and I encountered an error while checking out the model repo. So, I added an extra note about that.

Edit 2: despite having `$HF_TOKEN` set and being logged into Hugging Face (verified by `make setup`), I couldn't successfully run `download-models` due to an auth error. So I fixed the Makefile to include HF's username and token when cloning the model's repository (I'm not familiar with `huggingface-cli`, so I'm not sure how to properly fix this). Side note: users who have set an SSH key on the HF website shouldn't run into this auth issue.

ps. Thank you for open sourcing this project!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,8 @@ download-models:
 		cd $(MODEL_REPO_DIR) && git reset --hard origin/main; \
 	else \
 		echo "Repository not found, cloning..."; \
-		git clone git@hf.co:$(MODEL_REPO) $(MODEL_REPO_DIR); \
+		$(eval HF_USERNAME := $(shell huggingface-cli whoami | head -n 1)) \
+		git clone https://$(HF_USERNAME):$(HF_TOKEN)@hf.co/$(MODEL_REPO) $(MODEL_REPO_DIR); \
 	fi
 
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -34,7 +34,8 @@ download-models:
 		cd $(MODEL_REPO_DIR) && git reset --hard origin/main; \
 	else \
 		echo "Repository not found, cloning..."; \
-		git clone git@hf.co:$(MODEL_REPO) $(MODEL_REPO_DIR); \
+		$(eval HF_USERNAME := $(shell huggingface-cli whoami)) \
```
ZachNagengast, line 37:
```suggestion
		$(eval HF_USERNAME := $(shell huggingface-cli whoami  | head -n 1)) \
```

ZachNagengast (changes requested):
Awesome @thenameless7741 thanks for PR #1 🎉 
This looks like a great fix, just recommended a check in case the user is in an HF org (it shows up as a second line)

Could also be reduced to just 

`HF_USERNAME=$(huggingface-cli whoami | head -n 1) \`

What do you think?

Makefile
thenameless7741: Yes, your solution is more robust. I have updated the PR with this change.

ZachNagengast (approved):
Looks great, nicely done!

PR #8, opened 2024-02-01
Fix memory leak

finnvoor:
Without this I got up to 12gb of memory used when transcribing a 2hr audio file with base.en.


ZachNagengast:
Nice find! Testing this out

ZachNagengast (approved):
All the tests pass, good to merge 🚀

PR #18, opened 2024-02-12
Added implementation for SuppressBlankFilter

jkrukowski:
* added implementation for `SuppressBlankFilter`
* added tests

### DIFF
--- a/Sources/WhisperKit/Core/Utils.swift
+++ b/Sources/WhisperKit/Core/Utils.swift
@@ -14,22 +14,26 @@ import Tokenizers
 // MARK: - Helpers
 
 extension MLMultiArray {
-    /// Calculate the linear offset by summing the products of each dimension’s
-    /// index with the dimension’s stride
-    ///
-    /// More info here: https://developer.apple.com/documentation/coreml/mlmultiarray/2879231-subscript
-    func linearOffset(for index: [NSNumber]) -> Int {
+    /// Calculate the linear offset by summing the products of each dimension’s index with the dimension’s stride.
+    /// More info [here](https://developer.apple.com/documentation/coreml/mlmultiarray/2879231-subscript)
+    /// - Parameters:
+    ///  - index: The index of the element
+    ///  - strides: The precomputed strides of the multi-array, if not provided, it will be computed. It's a performance optimization to avoid recomputing the strides every time when accessing the multi-array with multiple indexes.
+    @inline(__always)
+    func linearOffset(for index: [NSNumber], strides strideInts: [Int]? = nil) -> Int {
         var linearOffset = 0
-        for (dimension, stride) in zip(index, strides) {
-            linearOffset += dimension.intValue * stride.intValue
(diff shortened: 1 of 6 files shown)


Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -525,7 +525,7 @@ public class WhisperKit {
 
                 // add them to the `allSegments` list
                 allSegments.append(contentsOf: currentSegments)
-                let allCurrentTokens = currentSegments.reduce([]) { $0 + $1.tokens }
+                let allCurrentTokens = currentSegments.flatMap { $0.tokens }
```
jkrukowski, line 528: I think `flatMap` will be more performant here

Sources/WhisperKit/Core/Utils.swift
```diff
@@ -25,6 +25,14 @@ extension MLMultiArray {
         }
         return linearOffset
     }
+
+    func fill<Value>(indexes: [[NSNumber]], with value: Value) {
+        let pointer = UnsafeMutablePointer<Value>(OpaquePointer(dataPointer))
+        for index in indexes {
```
ZachNagengast, line 31: This could take a while for large numbers of indexes - since this is a vector, it would be faster to parallelize this loop, here's one way to do it but open to other options (you could also use a taskgroup or similar) 
```suggestion
        DispatchQueue.concurrentPerform(iterations: indexes.count) { i in
            let index = indexes[i]
```

ZachNagengast (commented):
This looks great, appreciate the eye on performance. I made one suggestion below to speed it up a bit more, lmk your thoughts

Sources/WhisperKit/Core/Utils.swift
ZachNagengast: Reference to the task group option https://stackoverflow.com/a/72540228

jkrukowski: created a small benchmark here https://github.com/jkrukowski/FillBenchmark to compare `fill` and `fill` with `DispatchQueue.concurrentPerform`. Looks like there is no speed up for a release build when using the latter.

Implementing `fill` using task group would most likely require changing the interface (`filterLogits` would have to be `async`) + it still might not bring the benefits, see the discussion [here](https://forums.swift.org/t/taskgroup-and-parallelism/51039/4). Would love to hear your opinion about it

jkrukowski: thanks! Merged your PR and updated readme with the new results in https://github.com/jkrukowski/FillBenchmark
Updated this PR with the precomputed version of the `linearOffset` function and added `@inline(__always)` directive which makes it equally performant as a standalone version. Keeping a separate `linearOffset` function might be useful in future

ZachNagengast: `@inline(__always)` is neat! Agree it'll be useful to have around.

ZachNagengast (approved):
Look great, nicely done again!

PR #34, opened 2024-02-21
Allow setting a downloadBase so downloaded models are not forced into the user's Documents folder

jordibruin:


Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -177,7 +177,7 @@ public class WhisperKit {
 
         // If a local model folder is provided, use it; otherwise, download the model
         if let folder = modelFolder {
-            self.modelFolder = URL(fileURLWithPath: folder)
+            self.modelFolder = URL(fileURLWithPath: folder)
```
atiorh, line 180: Please remove the trailing whitespace :)

atiorh (approved):
Makes sense!

Sources/WhisperKit/Core/WhisperKit.swift
ZachNagengast, line 180:
```suggestion
            self.modelFolder = URL(fileURLWithPath: folder)
```

Should be able to commit this directly

ZachNagengast (approved):

ZachNagengast (approved):

PR #38, opened 2024-02-29
Support Word Timestamps

### LINKED ISSUE #2: Word level timestamps
Segment level timestamps look good, great work guys.

Are token level timestamps currently supported somehow, or on the roadmap?

ZachNagengast:
This is intended as an initial "functional" PR. Example code and usage guidelines will be coming as a fast-follow. Will also update the default models on huggingface to have the appropriate outputs. In the meantime, you can use this CLI script to test out the flow:

1. Download tiny.en (only one with alignment weights currently)
```bash
make download-model MODEL=tiny.en
```

2. Transcribe with `--word-timestamps` flag
```bash
swift run transcribe --word-timestamps \
 --model-path "Models/whisperkit-coreml/openai_whisper-tiny.en" \
--audio-path ~/Downloads/ted_60.wav \
--report \
--report-path ~/Downloads \
--verbose
```
Outputs the following json:

https://gist.github.com/ZachNagengast/f36a751bc68a3b5f2c41ada8bcc33746

Resolves #2

### DIFF
--- a/Sources/WhisperKit/Core/Models.swift
+++ b/Sources/WhisperKit/Core/Models.swift
@@ -3,13 +3,14 @@
 
 import CoreML
 import Hub
+import NaturalLanguage
 import Tokenizers
 
 #if os(watchOS) || arch(arm64)
-    @available(macOS 14, iOS 17, watchOS 10, visionOS 1, *)
-    public typealias FloatType = Float16
+@available(macOS 14, iOS 17, watchOS 10, visionOS 1, *)
+public typealias FloatType = Float16
 #else
-    public typealias FloatType = Float
+public typealias FloatType = Float
 #endif
 
 // MARK: - CoreML
@@ -54,37 +55,37 @@ public enum ModelVariant: CustomStringConvertible, CaseIterable {
     // TODO: implement config.json and generation_config.json parsing for models
     public var isMultilingual: Bool {
         switch self {
-        case .tiny, .base, .small, .medium, .large, .largev2, .largev3:
-            return true
-        case .tinyEn, .baseEn, .smallEn, .mediumEn:
-            return false
+            case .tiny, .base, .small, .medium, .large, .largev2, .largev3:
+                return true
+            case .tinyEn, .baseEn, .smallEn, .mediumEn:
+                return false
         }
     }
 
     public var description: String {
         switch self {
-        case .tiny:
-            return "tiny"
-        case .tinyEn:
-            return "tiny.en"
-        case .base:
-            return "base"
-        case .baseEn:
-            return "base.en"
-        case .small:
-            return "small"
-        case .smallEn:
-            return "small.en"
-        case .medium:
-            return "medium"
-        case .mediumEn:
-            return "medium.en"
-        case .large:
-            return "large"
-        case .largev2:
-            return "large-v2"
-        case .largev3:
-            return "large-v3"
+            case .tiny:
+                return "tiny"
+            case .tinyEn:
+                return "tiny.en"
+            case .base:
+                return "base"
+            case .baseEn:
+                return "base.en"
+            case .small:
+                return "small"
+            case .smallEn:
+                return "small.en"
+            case .medium:
+                return "medium"
+            case .mediumEn:
+                return "medium.en"
+            case .large:
+                return "large"
+            case .largev2:
+                return "large-v2"
+            case .largev3:
+                return "large-v3"
         }
     }
 }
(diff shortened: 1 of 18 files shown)


Sources/WhisperKit/Core/Models.swift
```diff
@@ -321,10 +339,20 @@ public struct TranscriptionSegment: Hashable, Codable {
...
     public var noSpeechProb: Float
+    public var words: [WordTiming]?
+}
+
+public struct WordTiming: Hashable, Codable {
+    var word: String
+    var tokens: [Int]
+    var start: Float
+    var end: Float
+    var probability: Float
```
finnvoor, line 355: can't actually access the timing at the moment 😅

```suggestion
    public var word: String
    public var tokens: [Int]
    public var start: Float
    public var end: Float
    public var probability: Float
```

Sources/WhisperKit/Core/SegmentSeeker.swift
```diff
@@ -149,9 +180,328 @@ public class SegmentSeeker: SegmentSeeking {
...
+        return (textIndices.reversed(), timeIndices.reversed())
+    }
+
+    func mergePunctuations(alignment: [WordTiming], prepended: String, appended: String) -> [WordTiming] {
+        var mergedAlignment = [WordTiming]()
+
+        // Merge prepended punctuations
+        for i in 1..<alignment.count {
+            let currentWord = alignment[i]
+            if i > 0, currentWord.word.starts(with: " "), prepended.contains(currentWord.word.trimmingCharacters(in: .whitespaces)) {
```
finnvoor, line 284: Getting a crash at `mergedAlignment[mergedAlignment.count - 1]` when `i == 1` and `mergedAlignment` is empty
```suggestion
            if i > 1, currentWord.word.starts(with: " "), prepended.contains(currentWord.word.trimmingCharacters(in: .whitespaces)) {
```

finnvoor: actually it seems like the right fix might be to change `for i in 1..<` to `for i in 0..<`, otherwise it drops the first token.

finnvoor:
It seems like `skipSpecialTokens` isn't respected when using word level timestamps, I still get a bunch of <|22.42|> timing tokens. Not sure if this is intentional or not.

finnvoor:
The word timings seem to be one word behind where they should be, I could be doing something wrong but just matching the WordTiming start and end to video + audio, the timing seems to be super precise but exactly 1 word behind.  You can see in this video if you unmute and watch the highlighted words.

https://github.com/argmaxinc/WhisperKit/assets/8284016/e73a2ce6-2760-499f-8a37-9bc3443ddcc9

Sources/WhisperKit/Core/SegmentSeeker.swift
ZachNagengast: Based on your video I think the merge strategy needs a little re-work, notice the contractions being attached to the next words and so forth. I'll include this for the time being but expect it to be much improved shortly.

ZachNagengast:
@finnvoor Thanks for giving this an early look, these comments are super helpful. I'm really impressed you were able to get that video output working, thanks for sharing - fascinating to see it with your overlay.

> It seems like `skipSpecialTokens` isn't respected when using word level timestamps, I still get a bunch of <|22.42|> timing tokens. Not sure if this is intentional or not.

Open question - I agree that skipSpecialTokens should remove them, but also wondering if they should just be removed by default? I.e. perhaps someone wants special tokens in the text responses, but not in the word timings. 

> The word timings seem to be one word behind where they should be, I could be doing something wrong but just matching the WordTiming start and end to video + audio, the timing seems to be super precise but exactly 1 word behind. You can see in this video if you unmute and watch the highlighted words.

This appears to be correct, will investigate. Also mentioned in another comment the punctuations for contractions are a little off too because it's being combined with the _next_ word instead of the current word. Will continue to refine this but I suspect these are related. Will report back soon.

ZachNagengast:
@finnvoor Just pushed a fix for some of the issues you reported.

Here's a short clip of the properly aligned word subtitles, as suspected they were 1 off previously.

This latest commit should also handle contractions much better. 

https://github.com/argmaxinc/WhisperKit/assets/1981179/0cf435a8-3843-4146-9284-21503979aa58

finnvoor:
looks much better now!

https://github.com/argmaxinc/WhisperKit/assets/8284016/3bfc1b79-8e01-4e2b-bd14-ecd86ca49d57

I can't imagine there's much use for having word level timings for special tokens, since they aren't really associated with time in audio. I think every use of Whisper I've seen has filtered out special tokens anyway

ZachNagengast:
@finnvoor Thanks for the feedback, these will no longer include special tokens. Also added some of the heuristics from the openai reference repo. I did notice that large-v3 was giving some pretty off results (lots of 0s length words) but v2 was fine, so something to keep in mind.

atiorh (approved):
Great work and ready to merge!
- All model versions in https://huggingface.co/datasets/argmaxinc/whisperkit-coreml support `--word-timestamps` now
- I reran quality evaluations and the output quality is unaffected by these changes.
- Performance regresses slightly but is expected and we will address that in the next release

PR #45, opened 2024-03-04
Added TimestampRulesFilter implementation

jkrukowski:
This PR adds implementation for `TimestampRulesFilter`. The implementation is based on https://github.com/openai/whisper/blob/master/whisper/decoding.py#L441

Couple of questions here @ZachNagengast:
- `sampleBegin` param passed to `TimestampRulesFilter` is 0, I think it might be incorrect. I compared it to the python implementation from the OpenAI repo and there this param is always greater or equal than 3 (and this makes sense, first 3 tokens are special tokens: 50258, 50259 and 50359 and AFAIK we don't want to supress them). If you run this code as is, some segments might be omited (because of the `sampleBegin` is 0, if you change it to 3, it should be ok).
- this implementation slows down the whole inference code, maybe you have some ideas how to optimize it?
- you mentioned that is has duplicated logic with `SegmentSeeker`, but I don't see it (AFAIK `TimestampRulesFilter` just supresses the token probabilities, while `SegmentSeeker` creates the whole segments). Could you please clarify?


Sources/WhisperKit/Core/LogitsFilter.swift
```diff
@@ -46,19 +47,158 @@ public class SuppressBlankFilter: LogitsFiltering {
...
+            shape: .vector(timeTokenCount, stride: 1)
+        )!
+
+        let timestampLogProb = BNNSNDArrayDescriptor.allocateUninitialized(
+            scalarType: FloatType.self,
+            shape: .vector(1, stride: 1)
+        )
+        defer { timestampLogProb.deallocate() }
+
+        try! BNNS.applyReduction(
```
jkrukowski, line 169: not sure about this force unwrapping here

Tests/WhisperKitTests/UnitTests.swift
```diff
@@ -245,7 +245,7 @@ final class UnitTests: XCTestCase {
         let options = DecodingOptions(usePrefillPrompt: true, withoutTimestamps: false)
 
         let result = try? await whisperKit!.transcribe(audioArray: multiWindowSamples, decodeOptions: options)
-        XCTAssertEqual(result?.segments.count, 2, "Expected 2 segments")
+        XCTAssertEqual(result?.segments.count, 3, "Expected 3 segments")
```
jkrukowski, line 248: enabled timestamps are causing more segments to appear

Sources/WhisperKit/Core/LogitsFilter.swift
```diff
@@ -46,19 +47,158 @@ public class SuppressBlankFilter: LogitsFiltering {
...
+        noTimestampsToken: Int,
+        timeTokenBegin: Int,
+        endToken: Int,
+        sampleBegin: Int,
+        maxInitialTimestampIndex: Int?
+    ) {
+        self.noTimestampsToken = noTimestampsToken
+        self.timeTokenBegin = timeTokenBegin
+        self.endToken = endToken
+        self.sampleBegin = 3 // FIXME: it should not be hardcoded value
```
jkrukowski, line 69: this needs to be fixed before merging

ZachNagengast: Yes, this breaks the EN models I believe

ZachNagengast:
@jkrukowski I push a small commit to measure the logit filtering time, here is what I'm getting for tiny with and without these new timestamp rules on the `jfk.wav` file:
With:
`[WhisperKit] - Logit Filtering:     192.41 ms /     28 runs (    6.87 ms/run) 37.78%`
Without:
`[WhisperKit] - Logit Filtering:       0.07 ms /     28 runs (    0.00 ms/run)  0.02%`


This is a bit high, it becomes especially noticeable with the tiny model. Something that is interesting is that only the first and last few tokens are slow (graph by chatgpt). This is for the jfk.wav

image
Hopefully this gives you some guidance on where to look for optimizations. And the majority of the slowdown is in this block of code:

```swift
            // timestamps have to appear in pairs, except directly before EOT; mask logits accordingly
            let sampledTokens = tokens[sampleBegin...]
            let lastWasTimestamp = sampledTokens.count >= 1 && sampledTokens.last! >= timeTokenBegin
            let penultimateWasTimestamp = sampledTokens.count < 2 || sampledTokens.dropLast().last! >= timeTokenBegin
            if lastWasTimestamp {
                if penultimateWasTimestamp {
                    // has to be non-timestamp
                    logits.fillLastDimension(indexes: timeTokenBegin..<logits.count, with: -FloatType.infinity)
                } else {
                    // cannot be normal text tokens
                    logits.fillLastDimension(indexes: 0..<endToken, with: -FloatType.infinity)
                }
            }
```

jkrukowski:
> @jkrukowski I push a small commit to measure the logit filtering time, here is what I'm getting for tiny with and without these new timestamp rules on the `jfk.wav` file: With: `[WhisperKit] - Logit Filtering: 192.41 ms / 28 runs ( 6.87 ms/run) 37.78%` Without: `[WhisperKit] - Logit Filtering: 0.07 ms / 28 runs ( 0.00 ms/run) 0.02%`
> 
> This is a bit high, it becomes especially noticeable with the tiny model. Something that is interesting is that only the first and last few tokens are slow (graph by chatgpt). This is for the jfk.wav
> 
> image Hopefully this gives you some guidance on where to look for optimizations. And the majority of the slowdown is in this block of code:
> 
> ```swift
>             // timestamps have to appear in pairs, except directly before EOT; mask logits accordingly
>             let sampledTokens = tokens[sampleBegin...]
>             let lastWasTimestamp = sampledTokens.count >= 1 && sampledTokens.last! >= timeTokenBegin
>             let penultimateWasTimestamp = sampledTokens.count < 2 || sampledTokens.dropLast().last! >= timeTokenBegin
>             if lastWasTimestamp {
>                 if penultimateWasTimestamp {
>                     // has to be non-timestamp
>                     logits.fillLastDimension(indexes: timeTokenBegin..<logits.count, with: -FloatType.infinity)
>                 } else {
>                     // cannot be normal text tokens
>                     logits.fillLastDimension(indexes: 0..<endToken, with: -FloatType.infinity)
>                 }
>             }
> ```

@ZachNagengast I've added more performant version of `fillLastDimension` function, seems like it's doing better, this is what I get for the release build on the `jfk.wav` file:

```
[WhisperKit] ---- Transcription Timings ----
[WhisperKit] Audio Load:              2.33 ms /      1 runs (    2.33 ms/run)  0.66%
[WhisperKit] Audio Processing:        0.11 ms /      1 runs (    0.11 ms/run)  0.03%
[WhisperKit] Mels:                   35.53 ms /      1 runs (   35.53 ms/run) 10.11%
[WhisperKit] Encoding:               13.39 ms /      1 runs (   13.39 ms/run)  3.81%
[WhisperKit] Matrices Init:           0.22 ms /      1 runs (    0.22 ms/run)  0.06%
[WhisperKit] Prefill:                 0.00 ms /      1 runs (    0.00 ms/run)  0.00%
[WhisperKit] Decoding:              239.40 ms /     28 runs (    8.55 ms/run) 68.15%
[WhisperKit] Non-inference:          61.25 ms /     28 runs (    2.19 ms/run) 17.43%
[WhisperKit] - Logit Filtering:       3.24 ms /     28 runs (    0.12 ms/run)  0.92%
[WhisperKit] - Sampling:             14.17 ms /     28 runs (    0.51 ms/run)  4.03%
[WhisperKit] - Kv Caching:            2.79 ms /     28 runs (    0.10 ms/run)  0.80%
[WhisperKit] - Word Timestamps:       0.00 ms /      0 runs (    0.00 ms/run)  0.00%
[WhisperKit] - Windowing:             0.08 ms /      1 runs (    0.08 ms/run)  0.02%
[WhisperKit] Fallbacks:               0.00 ms /      0 runs (    0.00 ms/run)  0.00%
[WhisperKit] Decoding Full Loop:    351.06 ms /     28 runs (   12.54 ms/run) 99.93%
```

ZachNagengast:
Much better! This looks in line with what I was seeing for those faster middle tokens previously. Think this is ready to come out of draft now?

jkrukowski:
> Much better! This looks in line with what I was seeing for those faster middle tokens previously. Think this is ready to come out of draft now?

good to hear this, 2 things are left:
1. `self.sampleBegin = 3 // FIXME: it should not be hardcoded value` -- not sure what value should I put there
2. force unwrapping in `sumOfProbabilityOverTimestampsIsAboveAnyOtherToken` maybe we should not force unwrap and return false gracefully, wdyt?

ZachNagengast:
> 1. self.sampleBegin = 3 // FIXME: it should not be hardcoded value -- not sure what value should I put there

PrefilledIndex is already being passed into this function, but I think actually it should use `intialPromptIndex`. A good test to add for accuracy on this would be similar to this one https://github.com/argmaxinc/WhisperKit/blob/e45dc0a056197c4a4ee3dabe9c604f48b150e519/Tests/WhisperKitTests/UnitTests.swift#L314 where you'd create a bunch of options that change this initialPromptIndex and make sure it's working properly.

> 2. force unwrapping in sumOfProbabilityOverTimestampsIsAboveAnyOtherToken maybe we should not force unwrap and return false gracefully, wdyt?

Besides the verbosity I think it's ok. If you want to be extra safe, you can wrap that whole part in a do catch and log an error similar to the sampling code.  I'm not sure all the scenarios where BNNS will throw, but returning false would just fallback to default behavior so no issues there.

Sources/WhisperKit/Core/LogitsFilter.swift
```diff
@@ -46,19 +47,200 @@ public class SuppressBlankFilter: LogitsFiltering {
...
+    let endToken: Int
     let sampleBegin: Int
-    let maxInitialTimestamp: Int?
+    let maxInitialTimestampIndex: Int?
+    let isModelMultilingual: Bool
 
-    public init(tokenizer: Tokenizer, sampleBegin: Int) {
-        // TODO: implement
-        fatalError("Not implemented: \(#function)")
+    public init(
```
ZachNagengast, line 62: This interface is complex in the same way the SegmentSeeker is, I believe most of these are in the tokenizer object but that would require passing this in.

Sources/WhisperKit/Core/Utils.swift
```diff
@@ -135,6 +144,10 @@ func tokenizerNameForVariant(_ variant: ModelVariant) -> String {
     return tokenizerName
 }
 
+func isModelMultilingual(logitsDim: Int?) -> Bool {
```
ZachNagengast, line 147: We have this already here: https://github.com/argmaxinc/WhisperKit/blob/cf75348b3bd0a51cfdffbfaa7553df6a5b4e6bdf/Sources/WhisperKit/Core/Models.swift#L60

Thoughts on combining them? I think checking the logitDims is more robust, perhaps it can be set on the model or the textdecoder on load here: https://github.com/argmaxinc/WhisperKit/blob/cf75348b3bd0a51cfdffbfaa7553df6a5b4e6bdf/Sources/WhisperKit/Core/WhisperKit.swift#L313

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -615,7 +615,7 @@ public class WhisperKit: Transcriber {
 
                 // add them to the `allSegments` list
                 allSegments.append(contentsOf: currentSegments)
-                let allCurrentTokens = currentSegments.reduce([]) { $0 + $1.tokens }
+                let allCurrentTokens = currentSegments.flatMap { $0.tokens }
```
ZachNagengast, line 618: 👍

ZachNagengast (approved):
Approving for the pre-release tests but curious your thoughts on the comments, can be a future PR.

Sources/WhisperKit/Core/LogitsFilter.swift
jkrukowski: you're right, it's complex. I didn't want to make it dependent on `Tokenizer` so it's decoupled and relatively easier to test. I can change it if you think otherwise

ZachNagengast: No problem, your logic makes sense, we may want a simple object like `SpecialTokens` in the future, and extend tokenizer with it, rather than just adding these index properties themselves as extensions.

PR #48, opened 2024-03-04
skip functional tests for models that are not downloaded.

metropol:
hi all, it seems that functional tests require two models to be downloaded: openai_whisper-tiny (because of `testRealTimeFactorTiny()`) and openai_whisper-large-v3 (because of `testInitLarge()` and `testRealTimeFactorLarge()`). This should probably be added to the CONTRIBUTING.md, however this PR does not make this change.

Another functional test, `testOutputAll()` - actually `allModelPaths()`, assumes that folders in `./Models/whisperkit-coreml`, which are created from the `make setup-model-repo`, contain actual models, while in fact only models explicitly downloaded with `make download-model MODEL=x` or `make download-models` are present.

`testOutputAll()` fails if not all models have been downloaded, which is not ideal. Running all tests for all models (takes a long time and) makes sense in some scenarios, but probably not to test some code changes. 

I've implemented some changes so that `allModelPaths()` only returns folders that contain downloaded models. This is done by checking if a proxy file (`MelSpectrogram.mlmodelc/coremldata.bin`) is a git lfs pointer file (starting with `version https://…`). If it is, then the folder is not included.

### DIFF
--- a/Examples/WhisperAX/WhisperAX/Views/ContentView.swift
+++ b/Examples/WhisperAX/WhisperAX/Views/ContentView.swift
@@ -813,7 +813,7 @@ struct ContentView: View {
                         localModels.append(model)
                     }
                     
-                    availableLanguages = whisperKit.tokenizer?.langauges.map { $0.key }.sorted() ?? ["english"]
+                    availableLanguages = whisperKit.tokenizer?.languages.map { $0.key }.sorted() ?? ["english"]
                     loadingProgressValue = 1.0
                     modelState = whisperKit.modelState
                 }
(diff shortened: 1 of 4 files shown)


metropol:
Moved the typo fix to its own branch (for a future PR).

ZachNagengast (commented):
This is a nice improvement, thank you!

I'd support fixing the typo in this PR, I just re-ran your commit and it passed all the tests, it may have just failed due to a failed api request. If you want to bring it back it lmk, otherwise I can merge this as is.

metropol:
It failed when building for VisionOS, which I don't have as a platform locally.
For the typo fix (langauges -> languages), as you prefer: I can add it back to this branch or create another PR just for it.

ZachNagengast:
@metropol whatever is more convenient for you, probably just reverting the commit that removes it would be simplest.

metropol:
@ZachNagengast  I've reverted the commit the removes the typo.

ZachNagengast (approved):
Good to merge 🚀

PR #51, opened 2024-03-06
Implement selecting input device

cgfarmer4:
Allow for selection of Core Audio device to be used with `AVAudioEngine`. 

Please note this will only work on `macOS`.  Not sure theres a cleaner way to do this besides the `#if os(macOS)`.

https://streamable.com/vv0zjb

### DIFF
--- a/Examples/WhisperAX/WhisperAX/Views/ContentView.swift
+++ b/Examples/WhisperAX/WhisperAX/Views/ContentView.swift
@@ -133,12 +133,7 @@ struct ContentView: View {
                 controlsView
             }
             .toolbar(content: {
-                ToolbarItemGroup {
-                    #if os(macOS)
-                    audioDevicesView
-                        .frame(minWidth: 100, maxWidth: .infinity)
-                    #endif
-                    
+                ToolbarItem {
                     Button {
                         let fullTranscript = formatSegments(confirmedSegments + unconfirmedSegments, withTimestamps: enableTimestamps).joined(separator: "\n")
                         #if os(iOS)
@@ -311,30 +306,6 @@ struct ContentView: View {
     }
 
     // MARK: - Controls
-    #if os(macOS)
-    var audioDevicesView: some View {
-        VStack {
-            if let audioDevices = audioDevices, audioDevices.count > 0 {
-                Picker("", selection: $selectedAudioInput) {
-                    ForEach(audioDevices, id: \.self) { device in
-                        Text(device.name).tag(device.name)
-                    }
-                }
-                .disabled(isRecording)
-            }
-        }
-        .onAppear {
-            audioDevices = AudioProcessor.getAudioDevices()
-            if let audioDevices = audioDevices, 
-                !audioDevices.isEmpty,
-                selectedAudioInput == "No Audio Input",
-                let device = audioDevices.first {
-                selectedAudioInput = device.name
-            }
-        }
-    }
-    #endif
-
     var controlsView: some View {
         VStack {
             basicSettingsView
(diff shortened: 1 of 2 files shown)


cgfarmer4:
Seeing a test fail but looks unrelated to my changes?

```
Test Suite 'All tests' started at 2024-03-05 20:21:19.203.
Test Suite 'whisperkitPackageTests.xctest' started at 2024-03-05 20:21:19.205.
Test Suite 'FunctionalTests' started at 2024-03-05 20:21:19.205.
Test Case '-[WhisperKitTests.FunctionalTests testAsyncImplementation]' started.
WhisperKit/WhisperKit.swift:192: Fatal error: Unexpectedly found nil while unwrapping an Optional value
```

Examples/WhisperAX/WhisperAX/Views/ContentView.swift
```diff
@@ -854,12 +886,27 @@ struct ContentView: View {
...
                     return
                 }
+                
+                #if os(macOS)
+                var deviceId: AudioDeviceID?
+                if self.selectedAudioInput != "No Audio Input",
+                   let devices = self.audioDevices,
+                   let device = devices.first(where: {$0.name == selectedAudioInput}) {
+                    deviceId = device.id
+                }
```
ZachNagengast, line 896:
```suggestion
                }
                #endif
```

Examples/WhisperAX/WhisperAX/Views/ContentView.swift
```diff
@@ -854,12 +886,27 @@ struct ContentView: View {
...
                         bufferEnergy = whisperKit?.audioProcessor.relativeEnergy ?? []
                     }
                 }
+                #else
+                try? audioProcessor.startRecordingLive() { _ in
+                    DispatchQueue.main.async {
+                        bufferEnergy = whisperKit?.audioProcessor.relativeEnergy ?? []
+                    }
+                }
+                #endif
```
ZachNagengast, line 909:
```suggestion
```

ZachNagengast (changes requested):
Very nice, thanks for the great demo video.

There may be a couple ways to clean up the various macos flags, it does seem to be a bit more than necessary:
1. Create a typealias for macos `typealias DeviceID = AudioDeviceID`
2. Use this to keep `startRecordingLive` a single interface for every OS
3. Then you can keep that inputDeviceId optional everywhere, and only have flags around the specific macos code that makes it non-nil.

Doing this should reduce the amount of repeated code inside the macos flags by a good amount, want to keep the forking logic per OS to a minimum (some is unavoidable)

As far as the UI - I'd like to see this selection inside the `controlsView`, potentially in between the reset and the settings buttons, although you'll need to make adjustments for the stream tab, which is slightly different than transcribe tab.

cgfarmer4:
Awesome, thanks for the feedback. Ill work through this over the next couple days!

cgfarmer4:
@ZachNagengast cleaned up the forks. Not that experienced in multi-platform so thanks for the suggestion. Less forks, cleaner code 💪🏻. Controls look like this now. Tried to figure out a way to do it without throwing off the balance and this seems to work ok. Happy to continue to iterate if you have more feedback.

image

Lastly, when I first plugged in my audio interface, I got a new exception from CoreAudio. `required condition is false: format.sampleRate == hwFormat.sampleRate`, rebooting my machine resolved this but not sure how much we care to guard against this here. Found some potential paths [here](https://stackoverflow.com/questions/66971504/i-got-crash-when-record-required-condition-is-false-format-samplerate-hwf) but they dont seem aligned with the need for `16k` output. Thoughts?

ZachNagengast (approved):
Looks great! Running the tests then good to merge once complete. For the transcribe tab in particular, I was thinking of having it centered between these buttons, but not a blocker - will check it out to see what works well before the next release.

ZachNagengast:
Regarding the error, we do need to iron out requesting 16khz sample rate for the devices that support it, but that should be in it's own PR. We have a converter that will deal with anything not in 16khz but it adds a small & avoidable latency.

cgfarmer4:
Nice! I can add it between the Transcribe buttons in a follow up PR, np.

Issue #13, opened 2024-02-03
Feature request: In example, provide delete and re-download options


pradeepb28:
Models are taking really good time to download. If the WiFI is OFF for a second it gets struck - it would be great if the example has two more options - Delete and Retry.

ZachNagengast:
Hi @pradeepb28 can you elaborate a bit on what retry would entail? Would that be similar to "Cancel"?

pradeepb28:
Retry is like Cancel the current download that was struck, delete the partially downloaded model from the folder and start re-downloading. @ZachNagengast

ZachNagengast:
Understood, that makes sense. Putting this on the roadmap, but would make a great contribution from the community if anyone reading this is interested!

ZachNagengast:
this is more or less implemented with #54, regarding the redownload part, the example app now checks for partially downloaded models and redownloads them if it is unable to load the models, so redownloading should be much less necessary. Cancelling a download is still a needed feature though, which swift-transformers will support.

PR #54, opened 2024-03-07
Add audio device selector to transcribe + take a stab at Delete/Retry models

cgfarmer4:
Addresses #13 + adds audio device view in both sections.

1. New delete model button
Screenshot 2024-03-06 at 9 32 03 PM

2. Restart download of model if connection lost
Screenshot 2024-03-06 at 9 49 04 PM

3. Move audio input selection between the buttons on transcribe.
Screenshot 2024-03-06 at 9 49 17 PM

### DIFF
--- a/Examples/WhisperAX/WhisperAX/Views/ContentView.swift
+++ b/Examples/WhisperAX/WhisperAX/Views/ContentView.swift
@@ -246,6 +246,16 @@ struct ContentView: View {
                             .progressViewStyle(CircularProgressViewStyle())
                             .scaleEffect(0.5)
                     }
+                    
+                    Button(action: {
+                        deleteModel()
+                    }, label: {
+                        Image(systemName: "trash")
+                    })
+                    .help("Delete model")
+                    .buttonStyle(BorderlessButtonStyle())
+                    .disabled(localModels.count == 0)
+                    .disabled(!localModels.contains(selectedModel))
 
                     #if os(macOS)
                     Button(action: {
@@ -293,6 +303,21 @@ struct ContentView: View {
                             Text(String(format: "%.1f%%", loadingProgressValue * 100))
                                 .font(.caption)
                                 .foregroundColor(.gray)
+                            
+                            if modelState == .downloading {
+                                Spacer()
+                                
+                                Button(action: {
+                                    resetState()
+                                    loadingProgressValue = 0.0
+                                    loadModel(selectedModel, redownload: true)
+                                    modelState = .loading
+                                }, label: {
+                                    Image(systemName: "arrow.uturn.forward.circle")
+                                })
+                                .help("Restart download")
+                                .buttonStyle(BorderlessButtonStyle())
+                            }
                         }
                         if modelState == .prewarming {
                             Text("Specializing \(selectedModel) for your device...\nThis can take several minutes on first load")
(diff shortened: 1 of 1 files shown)


Examples/WhisperAX/WhisperAX/Views/ContentView.swift
```diff
@@ -779,13 +812,35 @@ struct ContentView: View {
                 try await whisperKit.loadModels()
 
                 await MainActor.run {
+                    if !localModels.contains(model) {
```
cgfarmer4, line 815: local models doesnt refresh if you redownload a model without this

ZachNagengast:
Nice idea with the delete and retry! This is probably a good time to handle canceling the previous download task in resetState() because it's a static method. As-is, if you are in the middle of a download and tap the restart button, the previous download will continue and a new one will start. It may require something like this:

```swift
        let downloadTask = Task {
            folder = try await WhisperKit.download(variant: model, from: repoName, progressCallback: { progress in
                DispatchQueue.main.async {
                    loadingProgressValue = Float(progress.fractionCompleted) * specializationProgressRatio
                    modelState = .downloading
                }

                // Check for cancellation here
                try Task.checkCancellation()
            })
        }
```

or something similar to how the progressbar task gets cancelled:

```swift

                let progressBarTask = Task {
                    await updateProgressBar(targetProgress: 0.9, maxTime: 240)
                }

                // Prewarm models
                do {
                    try await whisperKit.prewarmModels()
                    progressBarTask.cancel()
                } catch {
                    print("Error prewarming models, retrying: \(error.localizedDescription)")
                    progressBarTask.cancel()
                    if !redownload {
                        loadModel(model, redownload: true)
                        return
                    } else {
                        // Redownloading failed, error out
                        modelState = .unloaded
                        return
                    }
                }
```

What do you think? This would also solve a few edge cases with the model selection, like if you select a new model while the previous one is downloading.

ZachNagengast:
Also played around with the mic selector placement, what do you think of this setup? Added a PR to your repo: https://github.com/cgfarmer4/WhisperKit/pull/1

image
image

cgfarmer4:
Beautiful, merged your PR and will take a look at cancelling the download tonight.

cgfarmer4:
I removed the retry, getting downloads to cancel is going to require a larger refactor of this function.

```
let downloadTask = Task {
            folder = try await WhisperKit.download(variant: model, from: repoName, progressCallback: { progress in
                DispatchQueue.main.async {
                    loadingProgressValue = Float(progress.fractionCompleted) * specializationProgressRatio
                    modelState = .downloading
                }

                // Check for cancellation here
                try Task.checkCancellation()
            })
        }
```

Going for this path, you get an error that `Mutation of captured var 'folder' in concurrently-executing code`.

Then I tried this which fixes the warning: `Mutation of captured var 'folder' in concurrently-executing code; this is an error in Swift 6` but doesnt properly cancel the download.

```
DispatchQueue.main.async {
    folder = downloadFolder
}
```

Next approach I tried was assigning the main `Task` similar to how `transcriptionTask` is assigned but that requires a pretty another refactor around the `do/catch` statements. 

How about I leave in the delete + your update and call it on this one? The download cancellation needs a rethink or Im totally missing something.

ZachNagengast:
No problem! We can get to that later on, but this is a good addition as-is 👍

ZachNagengast (approved):

PR #40, opened 2024-03-02
macOS 13 support

ZachNagengast:
Addresses: #9 #32

PR #56, opened 2024-03-07
Fix memory leak from non-async MLModel prediction

finnvoor:
After #40, all calls to `MLModel.prediction` were automatically switched from the async variant to the sync variant since they have the same name but async is only available on 14+.  The switch to the sync variant resulted in memory leaks due to resources allocated in `prediction` never getting released, so apps ran out of memory when transcribing a few minutes of audio. 

I added `MLModel.asyncPrediction`, which uses the async variation when available, which has no memory leak issues. On iOS 16 and macOS 13, the prediction is wrapped in a task which acts as an autoreleasepool. This should fix the issue on both 13 and 14+.

Tested on:
- iPhone 15 Pro (iOS 17.4)
- iPhone SE 2nd Gen (iOS 16.6)
- Macbook Pro (macOS 14.0)

### DIFF
--- a/Sources/WhisperKit/Core/AudioEncoder.swift
+++ b/Sources/WhisperKit/Core/AudioEncoder.swift
@@ -46,7 +46,7 @@ public class AudioEncoder: AudioEncoding, WhisperMLModel {
 
         try Task.checkCancellation()
 
-        let outputFeatures = try await model.prediction(from: modelInputs, options: MLPredictionOptions())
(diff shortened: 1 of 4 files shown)


ZachNagengast (approved):
Thanks for the submission and apologies for missing this. We'll need a set of benchmarks to find memory and speed regressions that should require passing before merges, will raise this issue.

Also, at a later time we will want to split up these large Util.swift and Models.swift into smaller files with appropriate grouping. 

Tests are passing and everything looks good here, clear to merge so you're not blocked, will also send out one more patch release with this and #58

Issue #36, opened 2024-02-27
Publish WhisperKit CLI on Homebrew


atiorh:
It would be great if `brew install whisperkit` just works and the WhisperKit CLI target on macOS could become an out-of-the-box real-time transcription utility.

jkrukowski:
I can give it a try

ZachNagengast:
Cool! This will ideally be a github action that runs on any new release and does a couple of things:

- Builds/tests the CLI in release mode
- Uploads the artifact 
- Submits a PR into the homebrew repo with the new version

The last part will require a github token, so we can coordinate when you get to that point. There are some existing [github actions](https://github.com/marketplace?category=&type=actions&verification=&query=homebrew+sort%3Apopularity-desc) which may be worth looking into.

jkrukowski:
@ZachNagengast I'm gonna take the following steps here

- will submit manually the below `WhisperKit` formula to homebrew core repo.

```ruby
class WhisperKit < Formula
  desc "Swift native on-device speech recognition with Whisper for Apple Silicon"
  homepage "https://github.com/argmaxinc/WhisperKit"
  url "https://github.com/argmaxinc/WhisperKit.git", tag: "v0.3.3", revision: "bfa357e897dadae0dca6314424bb119149e375d2"
  license "MIT"

  depends_on xcode: ["15.0", :build]

  def install
    system "swift", "build", "-c", "release", "--product", "transcribe", "--disable-sandbox"
    bin.install ".build/release/transcribe"
  end

  test do
    system "#{bin}/transcribe", "--help"
  end
end
```

it means that cli installation will be `brew install whisper-kit` and once installed user will be able to run it like this `transcribe [<options>] [<supress-tokens> ...]`

- once the formula is merged I'll create a GitHub action to update it on every `v<RELEASE_VERSION>` tag

Some questions here:

- you mentioned as well that it should "Upload the artifact", what did you mean by that?
- let me know if the formula looks ok for so I can submit it

ZachNagengast:
This plan sounds great @jkrukowski thanks for the research here
- Is there a way to make sure the install command is `brew install whisperkit` without the dash
- Your plan for the github action sounds good
- Due to unfamiliarity with homebrew I was unaware we could allow users to build the repo themselves, and assumed we'd need to compile and upload a pre-build version, so no action is needed if this works. 
- Is there a way to limit the formula to M1-M3 devices? Just curious

atiorh:
[...]

PR #68, opened 2024-03-11
WhisperKit CLI cleanup

jkrukowski:
Before taking on https://github.com/argmaxinc/WhisperKit/issues/36 I decided to do a little cleanup in the CLI

- moved command line arguments to a separate `WhisperKitArguments` struct
- extracted a separate `transcribe` subcommand
- changed executable name to `whisper-kit`

so rn in order to run the transcribe command the user has to type `swift run whisper-kit transcribe`
It's going to be convenient if we're going to add more subcommands in future (e.g. `swift run whisper-kit translate`)


jkrukowski:
> Good thinking in general with the cleanup, however this will break quite a bit of our [benchmark/eval](https://github.com/argmaxinc/whisperkittools/blob/95d178c1ff062f075f689c63da379cefb613c777/whisperkit/pipelines.py#L176-L190) code as is, because it calls this function from python as a shell command, and would require updating all of that as well. If this is a blocker for the homebrew PR then we should discuss further, otherwise it would be ideal to keep the interfaces the same.

Sounds fair. I could potentially add a PR in https://github.com/argmaxinc/whisperkittools repo to accommodate for these changes but I understand the concern. I reverted the renaming part, please take a 2nd look @ZachNagengast

ZachNagengast (approved):
Looks great, will merge once tests pass 💯

Sources/WhisperKitCLI/transcribe.swift
```diff
@@ -4,82 +4,32 @@
 import ArgumentParser
 import CoreML
 import Foundation
-
 import WhisperKit
 
-@available(macOS 13, iOS 16, watchOS 10, visionOS 1, *)
 @main
```
ZachNagengast, line 10: This is still needed in order to build properly based on the other checks elsewhere
```suggestion
@available(macOS 13, iOS 16, watchOS 10, visionOS 1, *)
@main
```

PR #69, opened 2024-03-11
Fix crash with mic device sample rate mismatch 

ZachNagengast:
Previously the inputNode was using the sample rate based on the hardware that the app was running on. When selecting a microphone that is connected to the device but has a different hardware sample rate, there was a crash occuring:

```
required condition is false: format.sampleRate == hwFormat.sampleRate
```

This updates that code to always use the correct sample rate for the selected microphone.

Note: This PR also includes some minor fixes for the example app and tests


Sources/WhisperKit/Core/AudioProcessor.swift
```diff
@@ -511,31 +513,31 @@ public extension AudioProcessor {
...
         }
 
-        guard let converter = AVAudioConverter(from: inputFormat, to: desiredFormat) else {
+        guard let converter = AVAudioConverter(from: nodeFormat, to: desiredFormat) else {
             throw WhisperError.audioProcessingFailed("Failed to create audio converter")
         }
 
         let bufferSize = AVAudioFrameCount(minBufferLength) // 100ms - 400ms supported
-        inputNode.installTap(onBus: 0, bufferSize: bufferSize, format: inputFormat) { [weak self] (buffer: AVAudioPCMBuffer, _: AVAudioTime) in
+        inputNode.installTap(onBus: 0, bufferSize: bufferSize, format: nodeFormat) { [weak self] (buffer: AVAudioPCMBuffer, _: AVAudioTime) in
```
mispellingError, line 540: Would passing `format: nil` here work?

PR #71, opened 2024-03-13
Add `Progress` to `WhisperKit`

finnvoor:
Currently there doesn't seem to be a way to get the current transcription progress to display. This adds a `Progress` to WhisperKit for easily displaying transcription progress in a ProgressView as seen in the modified WhisperAX.

The progress could probably be much more fine grained by updating in the decoder loop (and possibly factoring in model load time/anything else?), but this is a pretty good start I think.


finnvoor:
> This is cool, good idea! Although it looks a bit strange in streaming mode because the full audio length increases with every loop, can we just disable it for the streaming tab?

Didn't notice that, should be fixed now ✅

PR #72, opened 2024-03-14
Updated swift-transformers and tokenizer changes

jkrukowski:
In this PR:

- updated to use the 0.1.3 version of the `swift-transformers`
- added param in CLI -- tokenizer config download path

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -15,7 +15,7 @@ let package = Package(
             targets: ["WhisperKit"]
         ),
         .executable(
-            name: "whisper-kit",
+            name: "whisperkit-cli",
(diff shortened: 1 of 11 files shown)


Package.swift
```diff
@@ -15,12 +15,12 @@ let package = Package(
             targets: ["WhisperKit"]
         ),
         .executable(
-            name: "transcribe",
+            name: "whisper-kit",
```
jkrukowski, line 18: this is a bit unfortunate, looks like we cannot have `whisperkit` executable and `WhisperKit` module because the whole thing is not compiling, @ZachNagengast please advise

Issue #29, opened 2024-02-16
Language Detection


ZachNagengast:
Language detection here should be fairly simple with logits filters now, it will entail a single decoder pass and sample just the language tokens. However, this cannot be used when we are using a prefill prompt (i.e. forced decoder tokens) so that will need special handling.

#### References

Openai implementation: https://github.com/openai/whisper/blob/ba3f3cd54b0e5b8ce1ab3de13e32122d0d5f98ab/whisper/decoding.py#L19
WhisperKit inline todo: https://github.com/argmaxinc/WhisperKit/blob/228630c37e4ac1b1c95790d77f64058d317f8859/Sources/WhisperKit/Core/TextDecoder.swift#L300

Abhinay1997:
Can we mark this as done so it reflects in the project page ?

ZachNagengast:
Closing now that #78 is merged, well done!

PR #78, opened 2024-03-17
Language Detection

Abhinay1997:
PR for the issue #29.

* Adds `LanguageLogitsFilter` that suppresses non-language logits
* Implements `detectLanguage` in `TextDecoder` that does a single forward pass from `<startoftranscript>` to get the language token
* Adds tests for `LanguageLogitsFilter`

### DIFF
--- a/Sources/WhisperKit/Core/LogitsFilter.swift
+++ b/Sources/WhisperKit/Core/LogitsFilter.swift
@@ -76,6 +76,7 @@ public class LanguageLogitsFilter: LogitsFiltering {
         self.sampleBegin = sampleBegin
     }
 
+    // Retain the logits that correspond to language tokens and suppress non-language tokens
     public func filterLogits(_ logits: MLMultiArray, withTokens tokens: [Int]) -> MLMultiArray {
         guard tokens.count == sampleBegin else{
             return logits
--- a/Sources/WhisperKit/Core/LogitsFilter.swift
+++ b/Sources/WhisperKit/Core/LogitsFilter.swift
@@ -66,14 +66,16 @@ public class TimestampRulesFilter: LogitsFiltering {
 
 @available(macOS 13, iOS 16, watchOS 10, visionOS 1, *)
 public class LanguageLogitsFilter: LogitsFiltering {
-    let allLanguageTokens: [Int]
+    let allLanguageTokens: Set<Int>
     let logitsDim: Int
     let sampleBegin: Int
+    let nonLanguageTokenIndexes: [[NSNumber]]
 
     public init(allLanguageTokens: [Int], logitsDim: Int, sampleBegin: Int) {
-        self.allLanguageTokens = allLanguageTokens
+        self.allLanguageTokens = Set(allLanguageTokens)
         self.logitsDim = logitsDim
         self.sampleBegin = sampleBegin
+        self.nonLanguageTokenIndexes = LanguageLogitsFilter.getNonLanguageTokenIndexes(logitsDim: self.logitsDim, allLanguageTokens: self.allLanguageTokens)
     }
 
     // Retain the logits that correspond to language tokens and suppress non-language tokens
@@ -81,13 +83,11 @@ public class LanguageLogitsFilter: LogitsFiltering {
         guard tokens.count == sampleBegin else{
             return logits
         }
-        //get the non-language token indexes and suppress them
-        let nonLanguageTokenIndexes = self.getNonLanguageTokenIndexes(logitsDim: self.logitsDim, allLanguageTokens: self.allLanguageTokens)
         logits.fill(indexes: nonLanguageTokenIndexes, with: -FloatType.infinity)
         return logits
     }
     
-    private func getNonLanguageTokenIndexes(logitsDim: Int, allLanguageTokens: [Int]) -> [[NSNumber]]{
+    private static func getNonLanguageTokenIndexes(logitsDim: Int, allLanguageTokens: Set<Int>) -> [[NSNumber]]{
         var indexes: [[NSNumber]] = []
         for i in 0..<logitsDim{
             if !allLanguageTokens.contains(i){
--- a/Sources/WhisperKit/Core/LogitsFilter.swift
+++ b/Sources/WhisperKit/Core/LogitsFilter.swift
@@ -62,3 +62,37 @@ public class TimestampRulesFilter: LogitsFiltering {
         return logits
     }
 }
+
+
+@available(macOS 13, iOS 16, watchOS 10, visionOS 1, *)
+public class LanguageLogitsFilter: LogitsFiltering {
+    let allLanguageTokens: [Int]
+    let logitsDim: Int
+    let sampleBegin: Int
+
+    public init(allLanguageTokens: [Int], logitsDim: Int, sampleBegin: Int) {
+        self.allLanguageTokens = allLanguageTokens
+        self.logitsDim = logitsDim
+        self.sampleBegin = sampleBegin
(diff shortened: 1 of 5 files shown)


Sources/WhisperKit/Core/TextDecoder.swift
```diff
@@ -308,14 +316,101 @@ public class TextDecoder: TextDecoding, WhisperMLModel {
...
-        options decoderOptions: DecodingOptions,
+        sampler tokenSampler: TokenSampling,
+        options: DecodingOptions,
         temperature: FloatType
-    ) -> [DecodingResult] {
+    ) async throws -> [DecodingResult] {
         // TODO: implement
         // predict logits for 1 iteration with sot
         // 1. LanguageLogitsFilter for only language tokens
         // 2. GreedyTokenSampler for most likely language
```
ZachNagengast, line 326:
```suggestion
        // Predict logits for 1 iteration with sot
        // 1. LanguageLogitsFilter for only language tokens
        // 2. GreedyTokenSampler for most likely language
```

ZachNagengast (commented):
This is looking really good. There are some audio files already in the test bundle that you can use for spanish and japanese https://github.com/argmaxinc/WhisperKit/tree/main/Tests/WhisperKitTests/Resources It would be great to have an additional test that checks for accuracy with these similar to the other language tests https://github.com/argmaxinc/WhisperKit/blob/59cb8516c708e3e2f18198002600026b5a1135ca/Tests/WhisperKitTests/UnitTests.swift#L348

It would also be great to see some sort of performance comparison of using the filter vs not, for example you can cherry pick this commit in order to track overall filtering time https://github.com/argmaxinc/WhisperKit/pull/45/commits/ee95bb0789e277f45981b7027744958aea78eecc, ideally it is as small of an increase as possible to the overall pipeline

ZachNagengast:
Added a couple CI fixes here FYI @Abhinay1997 should be in a working state now

Abhinay1997:
Thanks for the feedback ! Yes, let me add tests for Spanish and Japanese. Will need to look into the filtering time !

Abhinay1997:
@ZachNagengast, for comparing the detection with and without logits filter, I'll add a `applyFilter` param to `detectLanguage` then ? To toggle it on and off.

Abhinay1997:
The filter adds only `2.9802322387695312e-06` seconds. Negligible imo

Sources/WhisperKit/Core/LogitsFilter.swift
```diff
@@ -62,3 +62,38 @@ public class TimestampRulesFilter: LogitsFiltering {
...
+    
+    private func getNonLanguageTokenIndexes(logitsDim: Int, allLanguageTokens: [Int]) -> [[NSNumber]]{
+        var indexes: [[NSNumber]] = []
+        for i in 0..<logitsDim{
+            if !allLanguageTokens.contains(i){
+                indexes.append([0, 0, i as NSNumber])
+            }
+        }
+        return indexes
+    }
```
jkrukowski, line 98: - maybe this can be precomputed in the class `init`?
- converting `allLanguageTokens` to set (instead of the array) might result in more perfomant code since `contains` for set should be O(1)

Abhinay1997: Agreed ! Let me do that !

Sources/WhisperKit/Core/TextDecoder.swift
```diff
@@ -308,14 +316,100 @@ public class TextDecoder: TextDecoding, WhisperMLModel {
...
+        logProbs = sampleResult.logProbs
+        
+        let samplingTime = Date().timeIntervalSince(samplingStartTime)
+        timings.decodingSampling += samplingTime
+        
+        var detectedLanguage = tokenizer.decode(tokens: [nextToken]).dropFirst(2).dropLast(2)
+        var decodingResult = DecodingResult.emptyResults
+        decodingResult.timings = timings
+        decodingResult.language = String(detectedLanguage)
+        return [decodingResult]
```
ZachNagengast, line 411: There's quite a bit of repeated code here that may get out of date with the main decoder loop, if you're up for it would be great to see an attempt at unifying them, but won't block merge.

Tests/WhisperKitTests/UnitTests.swift
```diff
@@ -563,6 +563,18 @@ final class UnitTests: XCTestCase {
...
+        let tokensFilter1 = LanguageLogitsFilter(allLanguageTokens: [2, 4, 6], logitsDim: 7, sampleBegin: 0)
+        let logits1 = try MLMultiArray.logits([0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7])
+        let result1 = tokensFilter1.filterLogits(logits1, withTokens: [])
+        XCTAssertEqual(result1.data(for: 2), [-.infinity, -.infinity, 0.3, -.infinity, 0.5, -.infinity, 0.7])
+        
+        let tokensFilter2 = LanguageLogitsFilter(allLanguageTokens: [2, 4, 6], logitsDim: 7, sampleBegin: 0)
+        let logits2 = try MLMultiArray.logits([0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7])
+        let result2 = tokensFilter2.filterLogits(logits2, withTokens: [1])
+        XCTAssertEqual(result2.data(for: 2), [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7])
+    }
```
ZachNagengast, line 577: Should test language detection using the full tiny pipeline tests here for Spanish and Japanese, similar to these tests https://github.com/argmaxinc/WhisperKit/blob/main/Tests/WhisperKitTests/UnitTests.swift#L348

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -646,12 +646,25 @@ public class WhisperKit: Transcriber {
...
+                    ).first
+                    currentDecodingOptions.language = languageDecodingResult?.language
+                }
 
                 decodingResult = try await textDecoder.decodeText(
                     from: encoderOutput,
                     using: decoderInputs,
                     sampler: tokenSampler,
-                    options: options,
+                    options: currentDecodingOptions,
```
ZachNagengast, line 667: Is this renaming unnecessary now?

ZachNagengast (changes requested):
Great progress, just need those last tests and cleanup and its good to go!

Sources/WhisperKit/Core/LogitsFilter.swift
Abhinay1997: This is done

Tests/WhisperKitTests/UnitTests.swift
Abhinay1997: Yes Zach, this is only for the logits filter test. The language detection test will be added near the tests for the other languages

Sources/WhisperKit/Core/TextDecoder.swift
Abhinay1997: Let me do it in a seperate PR ?

Sources/WhisperKit/Core/WhisperKit.swift
Abhinay1997: It would be needed as I have to pass the detected language to the decoder loop so it can skip the forward pass on language token.

Sources/WhisperKit/Core/TextDecoder.swift
ZachNagengast: Sure 👍

PR #80, opened 2024-03-19
Fix setupModels error handling

finnvoor:
Fixes a force unwrapping crash when model downloading fails


ZachNagengast (approved):
Seems reasonable, what scenarios are you seeing it fail to download? Low connectivity, or no connectivity?

finnvoor:
Yeah both of those cases failed

PR #84, opened 2024-03-21
WhisperKit init refactor

jkrukowski:
- added `WhisperMLModel` conformance, no more conditional casting
- `setupModels` -> static `setupModelFolder` which makes `modelFolder` non optional
- removed `download` param, I think it's not needed. When the `modelFolder` param is `nil` we download, otherwise we load the model from the directory

Possible future improvements:
- load models (`featureExtractor`, `audioEncoder` and `textDecoder`) in parallel?
- we could add `forceDownload` param which, in case `modelFolder` is provided will redownload

### DIFF
--- a/Sources/WhisperKit/Core/WhisperKit.swift
+++ b/Sources/WhisperKit/Core/WhisperKit.swift
@@ -24,10 +24,10 @@ public class WhisperKit: Transcriber {
     public var tokenizer: Tokenizer?
 
     /// Protocols
+    public var featureExtractor: any FeatureExtracting & WhisperMLModel
+    public var audioEncoder: any AudioEncoding & WhisperMLModel
+    public var textDecoder: any TextDecoding & WhisperMLModel
     public var audioProcessor: any AudioProcessing
-    public var featureExtractor: any FeatureExtracting
-    public var audioEncoder: any AudioEncoding
-    public var textDecoder: any TextDecoding
     public var logitsFilters: [any LogitsFiltering]
     public var segmentSeeker: any SegmentSeeking
 
@@ -50,9 +50,9 @@ public class WhisperKit: Transcriber {
     public let progress = Progress()
     
     /// Configuration
-    public var modelFolder: URL?
+    public var modelFolder: URL
     public var tokenizerFolder: URL?
-    private let useBackgroundDownloadSession: Bool
+    public let useBackgroundDownloadSession: Bool
 
     public init(
         model: String? = nil,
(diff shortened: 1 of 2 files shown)


Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -192,45 +190,34 @@ public class WhisperKit: Transcriber {
...
-    ) async throws {
+        useBackgroundDownloadSession: Bool
+    ) async throws -> URL {
         // Determine the model variant to use
         let modelVariant = model ?? WhisperKit.recommendedModels().default
-
         // If a local model folder is provided, use it; otherwise, download the model
-        if let folder = modelFolder {
-            self.modelFolder = URL(fileURLWithPath: folder)
-        } else if download {
```
ZachNagengast, line 218: This is needed for cased where someone wants to create the object but handle the downloading themselves. In fact I've been meaning to move this to the transcribe call so that the WhisperKit init is completely independent of any internet requests and purely sets up the object in all scenarios. This is important for folks that want to use the auxiliary functions of whisperkit by itself, without initializing models. Only when the developer shows intent to transcribe without any available models, it downloads as a fallback.

ZachNagengast: I do like the idea of modelFolder being non-optional, but we can't rely on the download function directly, we should build "future" model folder that download would generate from other swift-transformer functions

ZachNagengast (changes requested):
This looks good overall, just need to make the WhisperKit object not dependent on any URL requests - see comment.

Sources/WhisperKit/Core/WhisperKit.swift
jkrukowski: sounds good, seems more complex then I thought. Reverted back some changes, please take a 2nd look

xiangliangX: "Can the model now be used completely offline? It seems that loadTokenizer still needs to connect to hf (huggingface) to download.

xiangliangX:
Can the model now be used completely offline? It seems that loadTokenizer still needs to connect to hf (huggingface) to download.

ZachNagengast (changes requested):
I'd like to avoid requiring `WhisperMLModel` for these because we will soon be working with MLX models that will follow the same TextDecoding protocols, but won't have associated MLModels. Everything else looks good

jkrukowski:
> I'd like to avoid requiring `WhisperMLModel` for these because we will soon be working with MLX models that will follow the same TextDecoding protocols, but won't have associated MLModels. Everything else looks good

not much is left in this PR after this is removed so I'll close it for now

PR #85, opened 2024-03-21
CLI Task Handling

ZachNagengast:
This is necessary for translations from other languages into English (X->English).

It also adds logic so that if a language value is passed, it always uses the prefill prompt. This is necessary for translations from X->X.


atiorh (approved):
LGTM!

PR #87, opened 2024-03-22
Tokenizer refactor, tests cleanup

jkrukowski:
Based on a discussion with @ZachNagengast in https://github.com/argmaxinc/WhisperKit/pull/45 this PR introduces:

- changes to tokenizer:
  - new `SpecialTokens` struct
  - new protocol `WhisperTokenizer` which inherits from `Tokenizer` protocol and has `specialTokens` property
  - new struct `WhisperTokenizerWrapper` which implements `WhisperTokenizer` protocol and forwards tokenizer related methods
- `TextDecoding` has new `isModelMultilingual` property
- `SuppressBlankFilter` and `TimestampRulesFilter` depend on `SpecialTokens`
- tests cleanup
  - new assertions helper methods
  - new file `TestUtils`

### DIFF
--- a/Sources/WhisperKit/Core/Models.swift
+++ b/Sources/WhisperKit/Core/Models.swift
@@ -966,9 +966,9 @@ struct WhisperTokenizerWrapper: WhisperTokenizer {
         return false
     }
 
-    /// Tokenizes the given texta into individual words and associated tokens
-    /// - Parameter tokenIds: Array of tokens to split
-    /// - Returns: Tuple containing and array of the split words and and all tokens for each word
+    /// Decodes token ids into individual words and per-word subtokens
+    /// - Parameter tokenIds: Array of tokens to decode and then split 
+    /// - Returns: Tuple containing and array of the split words and all tokens for each word
     func splitToWordTokens(tokenIds: [Int]) -> (words: [String], wordTokens: [[Int]]) {
         let decodedWords = tokenizer.decode(tokens: tokenIds.filter { $0 < specialTokens.specialTokenBegin })
 
--- a/Sources/WhisperKit/Core/Models.swift
+++ b/Sources/WhisperKit/Core/Models.swift
@@ -836,52 +836,73 @@ public class TextDecoderCachePrefillOutput: MLFeatureProvider {
     }
 }
 
-// MARK: Tokenizer
+// MARK: SpecialTokens
 
-public extension Tokenizer {
-    var whitespaceToken: Int { convertTokenToId(" ") ?? Self.defaultWhitespaceToken }
-    var specialTokenBegin: Int { convertTokenToId("<|endoftext|>") ?? Self.defaultSpecialTokenBegin }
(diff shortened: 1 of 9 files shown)


Sources/WhisperKit/Core/Models.swift
```diff
@@ -836,52 +836,73 @@ public class TextDecoderCachePrefillOutput: MLFeatureProvider {
...
 
-    func splitTokensOnUnicode(tokens: [Int]) -> (words: [String], wordTokens: [[Int]]) {
-        let decodedFull = decode(tokens: tokens)
+public protocol WhisperTokenizer: Tokenizer {
+    var specialTokens: SpecialTokens { get }
+    
+    func splitToWordTokens(tokenIds: [Int]) -> (words: [String], wordTokens: [[Int]])
+}
+
+struct WhisperTokenizerWrapper: WhisperTokenizer {
```
pcuenca, line 884: Is there anything that could be done in `swift-transformers` to make this easier to extend?

Sources/WhisperKit/Core/Models.swift
```diff
@@ -945,118 +966,198 @@ public extension Tokenizer {
...
-        "panjabi": "pa",
-        "moldavian": "ro",
-        "moldovan": "ro",
-        "sinhalese": "si",
-        "castilian": "es",
-        "mandarin": "zh",
-    ] }
+    /// Tokenizes the given texta into individual words and associated tokens
+    /// - Parameter tokenIds: Array of tokens to split
+    /// - Returns: Tuple containing and array of the split words and and all tokens for each word
```
pcuenca, line 971:
```suggestion
    /// Decodes token ids into individual words and per-word subtokens
    /// - Parameter tokenIds: Array of tokens to decode and then split 
    /// - Returns: Tuple containing and array of the split words and all tokens for each word
```

Sources/WhisperKit/Core/Models.swift
jkrukowski: not sure if there is anything that can be done, any thoughts @ZachNagengast?

jkrukowski: we just wanted to have an efficient store for special tokens which are commonly used throughout the codebase. Extending `swift-transformers` `Tokenizer` protocol has its limitations -- only computed properties are allowed so I decided to go this way

pcuenca: Yes, I wonder if we should have used classes in some places :)

ZachNagengast: Special tokens would be nice to have directly from the config. We have one issue with the whisper tokenizer where they changed the no speech token to be <|nocaptions|> on some models and <|nospeech|> on others, and the way we're initializing right now is actually encoding this string, which is not resilient to that change.

ZachNagengast (approved):
This looks good to go, we should just be conscious of adding too much to this wrapper that would be better of in swift-transformers

Issue #24, opened 2024-02-16
Timestamp Rules Logits Filter


ZachNagengast:
Timestamp rules are helpful to more consistently find reliable timestamps during decoding. 

**_Important note:_** We have already brought over some of this logic into the [`SegmentSeeker`](https://github.com/argmaxinc/WhisperKit/blob/228630c37e4ac1b1c95790d77f64058d317f8859/Sources/WhisperKit/Core/SegmentSeeker.swift#L16) which runs at the end of a full decode loop to generate the segments. This feature will need to detangle any repeated logic between them.

#### References:

Openai implementation: https://github.com/openai/whisper/blob/ba3f3cd54b0e5b8ce1ab3de13e32122d0d5f98ab/whisper/decoding.py#L441-L505

jkrukowski:
I can take it

PR #88, opened 2024-03-23
Support distil whisper models

ZachNagengast:
Distil whisper models are a great addition to the standard openai models especially for smaller devices, and with [distil large v3](https://huggingface.co/distil-whisper/distil-large-v3
) out now, its a great time to add support for them.

Important note: This required some significant changes to how we search for models, specifically that we now require the full model name like "openai_whisper-large-v3" rather than just "large-v3" in order to use the convenience methods for downloading a model without doing it manually using the modelFolder param.

### Swift

`let pipe = try? await WhisperKit(model: "distil-whisper_distil-large-v3")`

or

`let pipe = try? await WhisperKit(model: "distil*large-v3")`

### CLI

`swift run whisperkit-cli transcribe --model "large-v3" --model-prefix "distil" --audio-path ~/your-audio.mp3`

### DIFF
--- a/Examples/WhisperAX/WhisperAX/Views/ContentView.swift
+++ b/Examples/WhisperAX/WhisperAX/Views/ContentView.swift
@@ -57,7 +57,7 @@ struct ContentView: View {
     @State private var currentFallbacks: Int = 0
     @State private var lastBufferSize: Int = 0
     @State private var lastConfirmedSegmentEndSeconds: Float = 0
-    @State private var requiredSegmentsForConfirmation: Int = 2
+    @State private var requiredSegmentsForConfirmation: Int = 4
     @State private var bufferEnergy: [Float] = []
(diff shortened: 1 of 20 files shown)


Examples/WhisperAX/WhisperAX/Views/ContentView.swift
```diff
@@ -57,7 +57,7 @@ struct ContentView: View {
     @State private var currentFallbacks: Int = 0
     @State private var lastBufferSize: Int = 0
     @State private var lastConfirmedSegmentEndSeconds: Float = 0
-    @State private var requiredSegmentsForConfirmation: Int = 2
+    @State private var requiredSegmentsForConfirmation: Int = 4
```
ZachNagengast, line 60: Updating this to account for #24 increasing the amount of segments generated overall

PR #99, opened 2024-04-02
Memory and Latency Regression Tests

Abhinay1997:
* Log memory during regression test
* Keep track of latency during regression test
* Summarize measurements and save for every N(=100) samples
* Dump measurements to a json for easy visualization.

Checklist:
- [x] code clean up
- [x] add more stats for latency tests
~~- [ ] add WER calculations in swift~~ Moved out of this PR. Plan to include it as part of EN Normalization PR
- [x] get processor info and add it to test output. See `Utils.swift`
- [x] use a longer audio file. > 1 hr for tests. Using [earnings22](https://huggingface.co/datasets/argmaxinc/earnings22-12hours/tree/main)
- [x] test dynamically on all models similar to `testOutputAll`

See [colab](https://colab.research.google.com/drive/1R9VfUkJWZVjYeUTkkBcrSqgAN-yrP5RJ?usp=sharing) for plots


ZachNagengast:
We also need a scheme to detect regressions for these tests on a chip-by-chip basis. We can store the data inside the test resources directly, or host them in a huggingface repo.

Files should be named something like this:
`Resources/Fixtures/RegressionTests/\(Process.processor)/\(modelName)/\(date).json`

All of these should be at or below 1.05x of the baseline data. I.e. a given regression test can be up to 5% slower than the baseline to account for various hardware fluctuations.
- Word error rate
- Peak memory
- Tokens per second

We don't yet have code to calculate WER from swift, but from python we use this metric: https://huggingface.co/spaces/evaluate-metric/wer


Should also have a test that runs all available models from the whisper-coreml repo, similar to `testOutputAll`, however I wouldn't expect this to run on the github runners due to the limited resources they have, it will just be run manually before releases.

Abhinay1997:
* WER calculations are detailed [here](https://github.com/jitsi/jiwer/blob/master/jiwer/process.py), we'll have to implement this in Swift.
* Chip/Processor info is possible to obtain. Shouldn't be an issue. I'll add that.
* Understood on the file name ! Let me do that.

Tests/WhisperKitTests/RegressionTests.swift
```diff
@@ -0,0 +1,199 @@
...
+    func testBaseEnglishPerformanceOverTime() async throws{
+        try await testAndMeasureModelPerformance(model: "base.en")
+    }
+    
+    func testLargeV2PerformanceOverTime() async throws{
+        try await testAndMeasureModelPerformance(model: "large-v2")
+    }
+    
+    func testLargeV2949PerformanceOverTime() async throws{
+        try await testAndMeasureModelPerformance(model: "large-v2_949")
```
ZachNagengast, line 156: These are all subject to change over time unfortunately, so these tests should actually search the full HF repo for all models available in there in order to run these tests. You can use this method to find all available models: https://github.com/argmaxinc/WhisperKit/blob/8564ce2bbec03649526678f51f4e249e9e052a6c/Sources/WhisperKit/Core/WhisperKit.swift#L131-L136

Abhinay1997: Updated the code to use this function, Zack

Abhinay1997:
You'll need to install (xcparse)[https://www.chargepoint.com/engineering/xcparse] via `brew install chargepoint/xcparse/xcparse`
```
xcodebuild clean build-for-testing -scheme whisperkit-Package -destination generic/platform=macOS | xcpretty
xcodebuild test -only-testing WhisperKitTests/RegressionTests -scheme whisperkit-Package -destination "platform=macOS,arch=arm64" -resultBundlePath ~/Downloads
xcparse attachments ~/Downloads/<latest_xc_result_file>.xcresult
```
Note: xcparse command above will output attachments as files in current directory

Tests/WhisperKitTests/FunctionalTests.swift
```diff
@@ -167,4 +167,5 @@ final class FunctionalTests: XCTestCase {
         }
         XCTAssertGreaterThan(transcriptionResult.text.count, 0)
     }
+
```
ZachNagengast, line 170:
```suggestion
```

ZachNagengast (commented):
Looks great! And low risk because its all new code. Needs just a bit of linting which I can run for you. 

Can you give a bit of guidance on how to test this? I.e. what commands to run and what the expected output should be?

Abhinay1997:
Do let me know how I can improve on the linting !

As for running it, the commands above should work, if not, you can manually run it from XCode and see the test attachments in the Xcode test result.

ZachNagengast (approved):

ZachNagengast:
> Do let me know how I can improve on the lining !

Will have linting rules setup soon, until then this is good to merge 👍

PR #115, opened 2024-04-13
Respect skipSpecialTokens option in the decodingCallback function

shawiz:
I want to show the live transcription process as the transcription happens, and don't want it to show special characters. Right now, regardless of how you set skipSpecialTokens in the decoder options, it will include special characters in the decodingCallback function under .text. I fixed that so it follows the skipSpecialTokens option. I've verified it in the WhisperAX app.

### DIFF
--- a/Sources/WhisperKit/Core/TextDecoder.swift
+++ b/Sources/WhisperKit/Core/TextDecoder.swift
@@ -621,7 +621,9 @@ open class TextDecoder: TextDecoding, WhisperMLModel {
                 }
 
                 // Prepare results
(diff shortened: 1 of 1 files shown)


atiorh:
Thanks @shawiz ! I will let @ZachNagengast  chime in. On a separate note, we are currently lowering the "Eager mode" from the app layer down to the library so it can be built on without extra user code. Just mentioning it now, in case you had to rewrite or copy-paste code to use it so far.

ZachNagengast (approved):
This looks great, thank you for the contribution!

shawiz:
Thanks guys!

PR #121, opened 2024-04-21
Fix audio file requirement for streaming mode

couche1:
When running cli with the `--stream` flag, an error occurs:
```
Either audioPath or audioFolder must be provided.
```

This PR resolves the issue by introducing a conditional to skip the audio file check in streaming mode.


ZachNagengast (approved):
Good catch, thank you!

PR #129, opened 2024-04-28
Added MLX feature extractor implementation

jkrukowski:
Based on https://github.com/ml-explore/mlx-examples/blob/c012eb173f0f632e369ec71f08be777df3aede08/whisper/whisper/audio.py#L130 implementation 

I've compared the outcome of the python implementation (based on first 30 seconds of Alice.mp3):
pyalice30

To this one:
alice30


Package.resolved
```diff
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/ml-explore/mlx-swift",
       "state" : {
-        "revision" : "7838f8cd93499f3d9e9a35b87cb74fe2664b325f",
-        "version" : "0.10.0"
+        "branch" : "main",
+        "revision" : "b43bdff8b6a413eb75e88eafd4a3995971a406fd"
```
jkrukowski, line 9: this is needed, otherwise `[.stride(by: -1)]` gives a compilation error

Sources/WhisperKit/MLX/MLXFeatureExtractor.swift
```diff
@@ -0,0 +1,51 @@
...
+        let result = try MLMultiArray(shape: logMelSpectrogram.shape.map { NSNumber(value: $0) }, dataType: .float32)
+        let maxRow = logMelSpectrogram.shape[0]
+        let maxCol = logMelSpectrogram.shape[1]
+        for row in 0..<maxRow {
+            for col in 0..<maxCol {
+                let key = [row, col] as [NSNumber]
+                result[key] = NSNumber(value: logMelSpectrogramData[row * maxCol + col])
+            }
+        }
+        return result
```
jkrukowski, line 49: maybe there is a faster way to create a `MLMultiArray` from `MLXArray`? we could ask the `MLX` team about it

Tests/WhisperKitTests/MLXTests.swift
```diff
@@ -8,15 +7,21 @@ import XCTest
...
+        let paddedSamples = try XCTUnwrap(
+            AudioProcessor.padOrTrimAudio(fromArray: audioSamples, startAt: 0, toLength: 480_000),
+            "Failed to pad audio samples"
+        )
+        let featureExtractor = MLXFeatureExtractor()
+        let melSpectrogram = try await XCTUnwrapAsync(
+            await featureExtractor.logMelSpectrogram(fromAudio: paddedSamples),
+            "Failed to produce Mel spectrogram from audio samples"
+        )
+        let expectedShape: [NSNumber] = [3000, 80]
```
jkrukowski, line 23: the output shape of the MLX feature extractor is different from the CoreML one

Sources/WhisperKit/MLX/MLXFeatureExtractor.swift
ZachNagengast: @davidkoski @awni Curious to hear your thoughts on this if you have any

awni: Check out [asData](https://swiftpackageindex.com/ml-explore/mlx-swift/main/documentation/mlx/mlxarray/asdata(nocopy:)), that might be useful here especially if you can avoid the copy.

Tests/WhisperKitTests/MLXTests.swift
atiorh: Yes, the existing Core ML Melspectrogram prepares its output for direct consumption by the Core ML AudioEncoder which anticipates `[1, {80,128}, 1, 3000]` (N=128 for large variants). We need to standardize the interface such that Core ML and MLX blocks are fungible

Sources/WhisperKit/MLX/MLXFeatureExtractor.swift
jkrukowski: came up with something like this

```swift
extension MLXArray {
    func asMLMultiArray(
        dataType: MLMultiArrayDataType = .float32,
        noCopy: Bool = true
    ) throws -> MLMultiArray {
        var data = asData(noCopy: noCopy)
        return try data.withUnsafeMutableBytes { (ptr: UnsafeMutableRawBufferPointer) in
            try MLMultiArray(
                dataPointer: ptr.baseAddress!,
                shape: shape.map { NSNumber(value: $0) },
                dataType: dataType,
                strides: strides.map { NSNumber(value: $0) }
            )
        }
    }
}
```

davidkoski: I wasn't sure about using Data as it is a struct and you are meant to use the pointer inside the callback.  I think something like this would probably work (it compiles but not tested):

```swift
        let array = MLXArray.ones([3, 4, 5])
        
        let coreMLType: MLMultiArrayDataType
        switch array.dtype {
        case .bool, .bfloat16, .complex64: fatalError("Unsupported type: \(array.dtype)")
        case .uint8, .uint16, .uint32, .uint64: fatalError("Unsupported type: \(array.dtype)")
        case .int8, .int16, .int64: fatalError("Unsupported type: \(array.dtype)")
        case .int32: coreMLType = .int32
        case .float16: coreMLType = .float16
        case .float32: coreMLType = .float32
        }
        
        // a buffer to be passed to CoreML
        let buffer = UnsafeMutableRawPointer.allocate(byteCount: array.nbytes, alignment: 8)
        
        // copy the data from the MLXArray backing into buffer
        array.asData(noCopy: true).withUnsafeBytes { ptr in
            let destination = UnsafeMutableRawBufferPointer(start: buffer, count: array.nbytes)
            ptr.copyBytes(to: destination)
        }
        
        func free(buffer: UnsafeMutableRawPointer) {
            buffer.deallocate()
        }
        
        let coreMLArray = try MLMultiArray(dataPointer: buffer, shape: array.shape.map { NSNumber(value: $0) }, dataType: coreMLType, strides: array.strides.map { NSNumber(value: $0) }, deallocator: free(buffer:))
```

jkrukowski: will look into it, thanks a lot @davidkoski

.github/workflows/unit-tests.yml
```diff
@@ -62,10 +62,10 @@ jobs:
...
-          xcodebuild test-without-building -only-testing WhisperKitTests/UnitTests -scheme whisperkit-Package -showdestinations
+          xcodebuild test-without-building -only-testing WhisperKitTests/UnitTests -scheme WhisperKit -showdestinations -skipPackagePluginValidation
       - name: Build and Test - ${{ matrix.run-config['name'] }}
         if: ${{ matrix.run-config['condition'] == true }}
         run: |
           set -o pipefail
-          xcodebuild clean build-for-testing -scheme whisperkit-Package -destination '${{ matrix.run-config['clean-destination'] }}' | xcpretty
-          xcodebuild test -only-testing WhisperKitTests/UnitTests -scheme whisperkit-Package -destination '${{ matrix.run-config['test-destination'] }}' | xcpretty
+          xcodebuild clean build-for-testing -scheme WhisperKit -destination '${{ matrix.run-config['clean-destination'] }}' -skipPackagePluginValidation | xcpretty
+          xcodebuild test -only-testing WhisperKitTests/UnitTests -scheme WhisperKit -destination '${{ matrix.run-config['test-destination'] }}' -skipPackagePluginValidation | xcpretty
```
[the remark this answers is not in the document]
jkrukowski: moved the MLX tests out of `WhisperKitTests`, so that `WhisperKitMLXTests` can be run separately

ZachNagengast (approved):
LGTM - pending after this is the watchOS investigation, and related, seeing how to get whisperkit-cli compiling without MLX

PR #130, opened 2024-04-29
Use IOSurface-backed MLMultiArrays for float16

smpanaro:
Copying large inputs, such as the KV cache, can add prediction latency for some device:OS combinations. On M1 Max and macOS Ventura this copying is ~25% of the prediction latency for whisper-large-v3. IOSurface-backed MLMultiArrays do not incur this copy.

Worth noting that Sonoma seems to have dramatically improved this from my testing.

Comparison from WhisperAX debug build using whisper-large-v3, M1 Max, macOS Ventura.
|Before|image|
|--|--|
|**After**|image|

### DIFF
--- a/Sources/WhisperKit/Core/Utils.swift
+++ b/Sources/WhisperKit/Core/Utils.swift
@@ -56,6 +56,24 @@ extension MLMultiArray {
             pointer[linearOffset] = value
         }
     }
+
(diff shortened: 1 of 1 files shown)


ZachNagengast (approved):
Confirmed this does give a speedup even on M3 Max on Sonoma (~5%), so looks like its a somewhat of a "free" speedup, and more noticeable on the older devices. Very nice find with this, thanks for the contribution 🎉

PR #136, opened 2024-05-08
Fix the reset operation exception in transcribeFile in the Demo.

jiangdi0924:
Clicking reset while operating transcribeFile in the Demo does not stop the transcribe  task.


ZachNagengast (approved):
This looks great, thanks for finding a nice simple solution!

PR #137, opened 2024-05-08
gh action for making pr to whisper-kit-expo on whisperkit release

seb-sep:


.github/workflows/expo-update.yml
```diff
@@ -0,0 +1,64 @@
...
+            git config --global user.email 'github-actions@github.com'
+
+        - name: New branch
+          run: |
+            git checkout -b $BRANCH_NAME
+            echo ${{ github.event.release }}
+            echo "Release tag is $TAG"
+        
+        - name: Run update script
+          run: node ./scripts/update-whisperkit.js $TAG
```
ZachNagengast, line 41: This is potentially a security issue without verifying what is in this js file ahead of time with a checksum -  the easiest would be to just have the code running directly here in the workflow. Here is a possible solution (please verify)
```suggestion
        - name: Update package.json version
          run: |
            PACKAGE_PATH="package.json"

            # Check if the package.json exists at the specified path
            if [ ! -f "$PACKAGE_PATH" ]; then
              echo "Could not find package.json at path: $PACKAGE_PATH."
              exit 1
            fi

            # Normalize the release tag by removing a leading 'v', if present
            RELEASE_TAG=${TAG#v}

            # Update package.json with the new version
            jq --arg newver "$RELEASE_TAG" '.whisperKit.version = $newver' "$PACKAGE_PATH" > tmp.$$.json && mv tmp.$$.json "$PACKAGE_PATH"

            # Verify changes
            cat "$PACKAGE_PATH"
```

.github/workflows/expo-update.yml
```diff
@@ -0,0 +1,64 @@
...
+            git push origin $BRANCH_NAME
+        
+        - name: Install and setup GitHub CLI if necessary
+          run: |
+            if ! command -v gh &> /dev/null
+            then
+              echo "GitHub CLI is not installed, installing..."
+              sudo apt update
+              sudo apt install -y gh
+              echo ${{ secrets.GITHUB_TOKEN }} | gh auth login --with-token
```
ZachNagengast, line 56:
```suggestion
              echo ${{ secrets.COMMITTER_TOKEN }} | gh auth login --with-token
```

ZachNagengast (changes requested):
Cool to see this, nice job! Left a few suggestions on how to use our bot user, please test them out and lmk.

.github/workflows/expo-update-test.json
```diff
@@ -0,0 +1,5 @@
+{
+    "release": {
+        "tag_name": "v0.6.1"
+    }
+}
```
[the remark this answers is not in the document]
seb-sep: When testing locally with `act`, the `github.event` object is null, which will not be the case in prod, so you can pass in a dummy event object to `act with `act -e event.json`. That is, this file is only used for local testing

ZachNagengast (approved):

Issue #142, opened 2024-05-22
how


zacfire:

jkrukowski:
In `WhisperKit` there are 2 methods with the same arguments but different return type

1. `func transcribe(audioPath:decodeOptions:callback:) async throws -> TranscriptionResult?`
2. `func transcribe(audioPath:decodeOptions:callback:) async throws -> [TranscriptionResult]`

The 1st one is deprecated and is subject to removal in a future version.

To suppress the warning, please specify the return type explicitly, e.g.

```swift
// ...
let whisper = try await WhisperKit()
let results: [TranscriptionResult] = try await whisper.transcribe(audioPath: audioPath)
// ...
```

ZachNagengast:
@zacfire Please pull the latest and try it out with the changes from #143 thanks to @jkrukowski

PR #143, opened 2024-05-22
Added `@_disfavoredOverload` for deprecated methods

jkrukowski:
Right now `WhisperKit` contains methods with the same arguments but different return type, e.g.

`func transcribe(audioPath:decodeOptions:callback:) async throws -> TranscriptionResult?`

and 

`func transcribe(audioPath:decodeOptions:callback:) async throws -> [TranscriptionResult]`

This results in the issue described here https://github.com/argmaxinc/WhisperKit/issues/142 where compiler gives an error `Ambiguous use of 'transcribe(audioPath:decodeOptions:callback:)`

This PR resolves that by adding `@_disfavoredOverload` to deprecated methods


ZachNagengast (approved):
Excellent use case for this, should address #142

PR #148, opened 2024-05-24
feat: add version output

### LINKED ISSUE #147: Add version support
It would be good to have version output for the CLI app

chenrui333:
```
$ swift build -c release --product whisperkit-cli --disable-sandbox
...

$ .build/arm64-apple-macosx/release/whisperkit-cli version
WhisperKitCLI version 0.7.0
```

fixes #147 

cc @ZachNagengast


ZachNagengast (commented):
Thanks for the help with the homebrew update. This is a great idea! Do you think there's a way to do this based on the latest git value or something automated? We're calling cli commands here 
https://github.com/argmaxinc/WhisperKit/blob/c829f9ada176737c3a60db79feea8a566b3f5196/Sources/WhisperKit/Core/Utils.swift#L122-L138
which seems to work well, perhaps current version could be a computed property based on `git describe --tags `

ZachNagengast (approved):
@chenrui333 Updated this PR with the built-in `version` param on `CommandConfiguration`. I've updated the formula https://github.com/argmaxincbot/homebrew-core/pull/1 such that it will use the tag from the URL to replace this placeholder, which will then be available in the build that homebrew makes. This way the version will always be up to date and self contained in the homebrew formula, lmk what you think

ZachNagengast:
Tests are passing, will merge if there's no objections in a bit.

chenrui333:
yeah, if we can make sure version is good as part of the release script, all good to me

PR #158, opened 2024-06-04
Resample audio files in 10mb chunks

### LINKED ISSUE #16: Resample audio file in chunks to reduce memory usage
https://github.com/argmaxinc/WhisperKit/blob/fed90c7c0727eb9f460d2a1226e0cbd5abf75141/Sources/WhisperKit/Core/AudioProcessor.swift#L197-L217

Creating an `AVAudioPCMBuffer` for the whole input audio buffer can easily surpass iOS memory limits.

Attempting to transcribe a 44100hz, 2 channel, ~1hr long video crashes on iOS due to running out of memory. It would be nice if instead of reading all the input audio into a buffer at once and converting,  the audio was read and converted in chunks to reduce the memory usage.

Another less common issue that would be solved by chunking the audio is that AVAudioPCMBuffer has a max size of UInt32.max, which can be hit when transcribing a 1-2hr, 16 channel, 44100hz audio file. This is a fairly typical audio file for a podcast recorded with a RODECaster Pro.

finnvoor:
closes #16 

Resampling audio files in 10mb chunks reduces the peak memory usage and fixes some niche issues with transcribing very long or very high sample rate / channel count audio files.

| Before | After |
|--------|--------|
| before | after | 

10mb is a bit arbitrary, but I chose it to roughly match the peak memory usage of the rest of the pipeline.

I expect this will have a very minor negative impact on speed of resampling, but given this is a small fraction of the time compared to the rest of the pipeline + the memory savings, it seems like a reasonable tradeoff.

### DIFF
--- a/Sources/WhisperKit/Core/AudioProcessor.swift
+++ b/Sources/WhisperKit/Core/AudioProcessor.swift
@@ -235,39 +235,45 @@ public class AudioProcessor: NSObject, AudioProcessing {
         }
 
         let frameCount = AVAudioFrameCount(audioFile.length)
-        guard let inputBuffer = AVAudioPCMBuffer(pcmFormat: audioFile.processingFormat, frameCapacity: frameCount),
+
+        // Read audio in 10mb increments to reduce the memory spike for large audio files
+        let maxReadFrameSize: AVAudioFrameCount = 10_000_000
+        guard let inputBuffer = AVAudioPCMBuffer(pcmFormat: audioFile.processingFormat, frameCapacity: min(frameCount, maxReadFrameSize)),
               let outputBuffer = AVAudioPCMBuffer(pcmFormat: outputFormat, frameCapacity: AVAudioFrameCount(newFrameLength))
         else {
             Logging.error("Unable to create buffers, likely due to unsupported file format")
             return nil
         }
 
-        do {
-            try audioFile.read(into: inputBuffer, frameCount: frameCount)
-        } catch {
-            Logging.error("Error reading audio file: \(error)")
-            return nil
-        }
-
-        let inputBlock: AVAudioConverterInputBlock = { _, outStatus in
-            if inputBuffer.frameLength == 0 {
-                outStatus.pointee = .endOfStream
+        while audioFile.framePosition < frameCount {
+            do {
+                let maxReadFrameCount = min(frameCount - UInt32(audioFile.framePosition), maxReadFrameSize)
+                try audioFile.read(into: inputBuffer, frameCount: maxReadFrameCount)
+            } catch {
+                Logging.error("Error reading audio file: \(error)")
                 return nil
-            } else {
-                outStatus.pointee = .haveData
-                return inputBuffer
             }
-        }
 
-        var error: NSError?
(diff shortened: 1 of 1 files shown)


ZachNagengast:
Amazing! Was just about to look at this, do you think there's any impact on audio quality a the breakpoints? I wouldn't expect much just curious

Sources/WhisperKit/Core/AudioProcessor.swift
```diff
@@ -235,39 +235,43 @@ public class AudioProcessor: NSObject, AudioProcessing {
         }
 
         let frameCount = AVAudioFrameCount(audioFile.length)
-        guard let inputBuffer = AVAudioPCMBuffer(pcmFormat: audioFile.processingFormat, frameCapacity: frameCount),
+        let chunkSize: AVAudioFrameCount = 10_000_000
+        guard let inputBuffer = AVAudioPCMBuffer(pcmFormat: audioFile.processingFormat, frameCapacity: min(frameCount, chunkSize)),
```
ZachNagengast, line 239: Nit, we use "chunk" often relating to the vad chunking, this will hopefully make it a little more distinct.
```suggestion

        // Read audio in 10mb increments to reduce the memory spike for large audio files
        let maxReadFrameSize: AVAudioFrameCount = 10_000_000
        guard let inputBuffer = AVAudioPCMBuffer(pcmFormat: audioFile.processingFormat, frameCapacity: min(frameCount, maxReadFrameSize)),
```

Sources/WhisperKit/Core/AudioProcessor.swift
```diff
@@ -235,39 +235,43 @@ public class AudioProcessor: NSObject, AudioProcessing {
...
-            Logging.error("Error reading audio file: \(error)")
-            return nil
-        }
-
-        let inputBlock: AVAudioConverterInputBlock = { _, outStatus in
-            if inputBuffer.frameLength == 0 {
-                outStatus.pointee = .endOfStream
+        while audioFile.framePosition < frameCount {
+            do {
+                let chunkedFrameCount = min(frameCount - UInt32(audioFile.framePosition), chunkSize)
```
ZachNagengast, line 248:
```suggestion
                let maxReadFrameCount = min(frameCount - UInt32(audioFile.framePosition), maxReadFrameSize)
```

ZachNagengast (approved):
Very nice @finnvoor thanks for the contribution! Next step will be purging the already transcribed audio during decoding.

finnvoor:
> Amazing! Was just about to look at this, do you think there's any impact on audio quality a the breakpoints? I wouldn't expect much just curious

hmm, not really sure but I doubt it would be enough to notice. I didn't test much but I got the same transcript when a file was split into ~16 chunks.

atiorh:
Thanks for the contrib @finnvoor! We will run full evals for 1.0.0 on all this behavior and address regressions (if any). This looks to be low risk but we might need to couple this with VAD to be double sure.

ZachNagengast:
FYI there appears to be an issue with this code that is placing audio in the wrong position in the outputBuffer. I am working on an approach that appends to the buffer every 10MB instead of writes directly to it.

Issue #100, opened 2024-04-02
Indeterminate tests


ZachNagengast:
For some reason, occasionally visionOS unit tests are failing at the very end, see: https://github.com/argmaxinc/WhisperKit/actions/runs/8525582362/job/23352857267?pr=99

The only error visible is `Error: The operation was canceled.` Would be great to figure out why that's happening and resolve it so the tests succeed properly.

jkrukowski:
Looks like here the reason is a timeout, see here https://github.com/argmaxinc/WhisperKit/actions/runs/8525582362?pr=99

<img width="1087" alt="Screenshot 2024-04-15 at 13 49 28" src="https://github.com/argmaxinc/WhisperKit/assets/487331/ddb6725d-16bc-441d-ab47-1ffc99232372">

unit tests on CI seems to be configured that way https://github.com/argmaxinc/WhisperKit/blob/8564ce2bbec03649526678f51f4e249e9e052a6c/.github/workflows/unit-tests.yml#L17

Should we increase the timeout (to e.g. 30 minutes)?

ZachNagengast:
I see, makes sense. In this case, we should not only run our tests in parallel but also run each platform in parallel. If its taking longer than 20 minutes that's too long for unit tests imo.

jkrukowski:
> I see, makes sense. In this case, we should not only run our tests in parallel but also run each platform in parallel. If its taking longer than 20 minutes that's too long for unit tests imo.

ok, let me try to do it then

ZachNagengast:
This issue appears to have started coming back, but for any test besides macOS

jkrukowski:
@ZachNagengast could you give a link to a failing job (somehow I couldn't find a relevant one), I'll try to inspect it

ZachNagengast:
@jkrukowski here is an example https://github.com/argmaxinc/WhisperKit/actions/runs/9543240677/job/26299468908?pr=166

PR #177, opened 2024-06-28
Fix TEST_HOST name mismatch

CongLeSolutionX:
Fix TEST_HOST name mismatch.

I saw this error message in the testing section, so I created this PR and submitted my fix. Hope you guys like it.

Before


ZachNagengast (approved):
Good catch, thank you!

CongLeSolutionX:
> Good catch, thank you!

I noticed a failed test in the pipeline. Let me investigate the issue, find a solution, and include it in this PR. Stay tuned!

ZachNagengast:
The test are currently somewhat indeterminate #100, so this is good to merge now.

PR #183, opened 2024-07-06
Fix resampling large files

### LINKED ISSUE #172: Only translating last 30s or so of the audio file.
When using the whisper-kit cli or apps with a large file, 50 minutes of audio, it looks like the final report (.srt file) is only showing the last 30s of content transcribed. 

Is this expected, Am I'm missing a command line argument?

ZachNagengast:
Resolves #172


atiorh (commented):
Leaving a review to trigger tests

PR #187, opened 2024-07-21
MLX Cleanup

jkrukowski:
In this PR:

- added new params to `WhisperKit`: 
  - `mlxModel` if specified, it will be used to specify which mlx model to download
  - `mlxModelRepo` if specified, it will be used to specify the repo where the mlx model is stored
  - `mlxModelFolder` if specified, it will point to the local folder where the mlx model is stored

- added new options to `CLIArguments`:
  - `mlxModel`, `mlxModelPath` see the explanation above
  - `featureExtractorType`, `audioEncoderType`, `textDecoderType` to specify which type (coreml or or mlx) should be used for given pipeline step

- updated readme to reflect the changes above
- updated makefile to reflect the changes above
- simplified CI unit test

To do:
- update MLX to the latest version, right now blocked by https://github.com/ml-explore/mlx-swift/issues/117, edit: fixed
- update the way we build brew, we should use `xcodebuild` instead of `swift build`

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -64,7 +64,7 @@ jobs:
           path: Models
           key: ${{ runner.os }}-models
       - name: Download Models
-        if: steps.model-cache.outputs.cache-hit != 'true'
+        # if: steps.model-cache.outputs.cache-hit != 'true'
         run: |
           make download-model MODEL=tiny 
           make download-mlx-model MODEL=tiny
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -31,7 +31,7 @@ jobs:
               test-destination: "platform=iOS Simulator,OS=${{ inputs.ios-version }},name=iPhone 15",
               test-cases: "-only-testing WhisperKitTests/UnitTests",
               mlx-disabled: "1",
-              scheme: "whisperkit",
+              scheme: "whisperkit-Package",
             }
           - {
               name: "watchOS",
@@ -39,7 +39,7 @@ jobs:
               test-destination: "platform=watchOS Simulator,OS=10.5,name=Apple Watch Ultra 2 (49mm)",
               test-cases: "-only-testing WhisperKitTests/UnitTests",
               mlx-disabled: "1",
-              scheme: "whisperkit",
+              scheme: "whisperkit-Package",
             }
           - {
               name: "visionOS",
@@ -47,7 +47,7 @@ jobs:
               test-destination: "platform=visionOS Simulator,name=Apple Vision Pro",
               test-cases: "-only-testing WhisperKitTests/UnitTests",
               mlx-disabled: "1",
-              scheme: "whisperkit",
+              scheme: "whisperkit-Package",
             }
     timeout-minutes: 30
     steps:
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -75,6 +75,8 @@ jobs:
           echo "Available schemes:"
           xcodebuild -list
           xcodebuild -downloadAllPlatforms
+          echo "Destinations for testing:"
+          export ${{ matrix.run-config['compiler-flags'] }} && xcodebuild test-without-building -only-testing WhisperKitTests/UnitTests -scheme ${{ matrix.run-config['scheme'] }} -showdestinations -skipPackagePluginValidation
       - name: Boot Simulator and Wait
         if: ${{ matrix.run-config['name'] != 'macOS' }} && ${{ inputs.macos-runner == 'macos-14' }}
         # Slower runners require some time to fully boot the simulator
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -61,7 +61,7 @@ jobs:
         id: model-cache
         uses: actions/cache@v4
         with:
-          path: Sources/WhisperKitTestsUtils/Models
+          path: Models
           key: ${{ runner.os }}-models
       - name: Download Models
         if: steps.model-cache.outputs.cache-hit != 'true'
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -20,7 +20,6 @@ jobs:
           - {
               name: "macOS",
               condition: true,
-              clean-destination: "generic/platform=macOS",
               test-destination: "platform=macOS,arch=arm64",
               test-cases: "-only-testing WhisperKitTests/UnitTests -only-testing WhisperKitMLXTests/MLXUnitTests",
               mlx-disabled: "0",
@@ -29,7 +28,6 @@ jobs:
           - {
               name: "iOS",
               condition: true,
-              clean-destination: "generic/platform=iOS",
               test-destination: "platform=iOS Simulator,OS=${{ inputs.ios-version }},name=iPhone 15",
               test-cases: "-only-testing WhisperKitTests/UnitTests",
               mlx-disabled: "1",
@@ -38,7 +36,6 @@ jobs:
           - {
               name: "watchOS",
               condition: "${{ inputs.macos-runner == 'macos-14' }}",
-              clean-destination: "generic/platform=watchOS",
               test-destination: "platform=watchOS Simulator,OS=10.5,name=Apple Watch Ultra 2 (49mm)",
               test-cases: "-only-testing WhisperKitTests/UnitTests",
               mlx-disabled: "1",
@@ -47,7 +44,6 @@ jobs:
           - {
               name: "visionOS",
               condition: "${{ inputs.macos-runner == 'macos-14' }}",
-              clean-destination: "generic/platform=visionOS",
               test-destination: "platform=visionOS Simulator,name=Apple Vision Pro",
               test-cases: "-only-testing WhisperKitTests/UnitTests",
               mlx-disabled: "1",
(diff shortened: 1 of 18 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -96,5 +90,8 @@ jobs:
         if: ${{ matrix.run-config['condition'] == true }}
         run: |
           set -o pipefail
-          xcodebuild clean build-for-testing -scheme ${{ matrix.run-config['scheme'] }} -destination '${{ matrix.run-config['clean-destination'] }}' -skipPackagePluginValidation | xcpretty
-          xcodebuild test -only-testing WhisperKitTests/UnitTests -scheme ${{ matrix.run-config['scheme'] }} -destination '${{ matrix.run-config['test-destination'] }}' -skipPackagePluginValidation
+          xcodebuild clean build-for-testing test \
+            ${{ matrix.run-config['test-cases'] }} \
+            -scheme ${{ matrix.run-config['scheme'] }} \
+            -destination '${{ matrix.run-config['test-destination'] }}' \
+            -skipPackagePluginValidation
```
jkrukowski, line 97: this is merged into single command `xcodebuild clean build-for-testing test`

Makefile
```diff
@@ -111,19 +112,38 @@ download-mlx-model: setup-mlx-model-repo
...
 build:
 	@echo "Building WhisperKit..."
-	@swift build -v
+	@xcodebuild CLANG_ENABLE_CODE_COVERAGE=NO VALID_ARCHS=arm64 clean build \
+		-configuration Release \
+		-scheme whisperkit-Package \
+		-destination generic/platform=macOS \
+		-derivedDataPath .build/.xcodebuild/ \
+		-clonedSourcePackagesDirPath .build/ \
+		-skipPackagePluginValidation
```
jkrukowski, line 124: we have to use `xcodebuild` because of `mlx`

Makefile
```diff
@@ -111,19 +112,38 @@ download-mlx-model: setup-mlx-model-repo
...
 
 test:
 	@echo "Running tests..."
-	@swift test -v
+	@xcodebuild clean build-for-testing test \
+		-scheme whisperkit-Package \
+		-only-testing WhisperKitMLXTests/MLXUnitTests \
+		-only-testing WhisperKitTests/UnitTests \
+		-destination 'platform=macOS,arch=arm64' \
+		-skipPackagePluginValidation
```
jkrukowski, line 145: we have to use `xcodebuild` because of `mlx`

Package.swift
```diff
@@ -53,12 +52,11 @@ func dependencies() -> [PackageDescription.Package.Dependency] {
...
 
 func mlxDependencies() -> [PackageDescription.Package.Dependency] {
-    let isMLXDisabled = ProcessInfo.processInfo.environment["MLX_DISABLED"] == "1"
-    if isMLXDisabled {
+    if isMLXDisabled() {
         return []
     } else {
         return [
-            .package(url: "https://github.com/ml-explore/mlx-swift", branch: "main"),
+            .package(url: "https://github.com/ml-explore/mlx-swift", revision: "d6d9472da5bf7ec2654e8914bd1d15622f45b6a9"),
```
jkrukowski, line 59: the latest version of `mlx` 0.16.0 breaks the implementation, let's pin it to the last commit before the breaking one, I'll have to talk to `mlx` team about the issue

README.md
```diff
@@ -80,76 +80,80 @@ This example demonstrates how to transcribe a local audio file:
...
-Task {
-   let pipe = try? await WhisperKit()
-   let transcription = try? await pipe!.transcribe(audioPath: "path/to/your/audio.{wav,mp3,m4a,flac}")?.text
-    print(transcription)
-}
+let pipe = try await WhisperKit()
+// Transcribe the audio file
+let transcription = try await pipe.transcribe(audioPath: "path/to/your/audio.{wav,mp3,m4a,flac}")?.text
+// Print the transcription
+print(transcription)
```
jkrukowski, line 87: removed not needed `Task`, decided to remove `?` from `try ?`. When the potential user does copy paste from the readme I think it's better when the error is not ignored

Package.swift
jkrukowski: fixed, solution provided here https://github.com/ml-explore/mlx-swift/issues/117

Makefile
```diff
@@ -7,9 +7,9 @@ PYTHON_COMMAND := python3
 MODEL_REPO := argmaxinc/whisperkit-coreml
 MLX_MODEL_REPO := argmaxinc/whisperkit-mlx
 
-MODEL_REPO_DIR := ./Models/whisperkit-coreml
-MLX_MODEL_REPO_DIR := ./Models/whisperkit-mlx
-BASE_MODEL_DIR := ./Models
+MODEL_REPO_DIR := ./Sources/WhisperKitTestsUtils/Models/whisperkit-coreml
+MLX_MODEL_REPO_DIR := ./Sources/WhisperKitTestsUtils/Models/whisperkit-mlx
+BASE_MODEL_DIR := ./Sources/WhisperKitTestsUtils/Models
```
ZachNagengast, line 12: Would be preferred to keep these models at the top level.

README.md
```diff
@@ -79,77 +79,87 @@ This example demonstrates how to transcribe a local audio file:
...
 ```
 
 ### Model Selection
 
-WhisperKit automatically downloads the recommended model for the device if not specified. You can also select a specific model by passing in the model name:
+You have to specify the model by passing the model name:
```
ZachNagengast, line 92: Why this change?

README.md
```diff
@@ -79,77 +79,87 @@ This example demonstrates how to transcribe a local audio file:
...
 
 ```bash
-swift run whisperkit-cli transcribe --model-path "Models/whisperkit-coreml/openai_whisper-large-v3" --audio-path "path/to/your/audio.{wav,mp3,m4a,flac}" 
+make setup
+make download-model MODEL=tiny
+make download-mlx-model MODEL=tiny
 ```
 
-Which should print a transcription of the audio file. If you would like to stream the audio directly from a microphone, use:
+and then run the tests:
```
ZachNagengast, line 159: Seems like some useful stuff got deleted in this section based on the diff here.

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -211,114 +228,80 @@ open class WhisperKit {
...
-
     public func prewarmModels() async throws {
         try await loadModels(prewarmMode: true)
     }
 
     public func loadModels(
         prewarmMode: Bool = false
     ) async throws {
-        modelState = prewarmMode ? .prewarming : .loading
+        assert(modelFolder != nil || mlxModelFolder != nil, "Please specify `modelFolder` or `mlxModelFolder`")
```
ZachNagengast, line 238: Will this crash on prod builds? If so it should throw instead of crash.

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -211,114 +228,80 @@ open class WhisperKit {
         }
     }
 
-    /// Sets up the model folder either from a local path or by downloading from a repository.
-    public func setupModels(
```
ZachNagengast, line 215: Is this deleted for a specific reason? Would be preferred to keep with with some adjustments since it was previously a public function and may be getting called elsewhere.

README.md
```diff
@@ -79,77 +79,87 @@ This example demonstrates how to transcribe a local audio file:
...
 
 Note that the model search must return a single model from the source repo, otherwise an error will be thrown.
 
 For a list of available models, see our [HuggingFace repo](https://huggingface.co/argmaxinc/whisperkit-coreml).
+For MLX models, see [here](https://huggingface.co/argmaxinc/whisperkit-mlx).
+
+If you want to get the recommended model for your device, you can use the following method:
+
+```swift
+print(WhisperKit.recommendedModel())
```
ZachNagengast, line 112:
```suggestion
print(WhisperKit.recommendedModels())
```

Would be useful to have an example of what this outputs as well.

ZachNagengast (changes requested):
Looking great, in the final stages of completing the MLX support project 🎉. Left a few comments and questions here to consider.

Sources/WhisperKit/Core/WhisperKit.swift
jkrukowski: it's not going to crash on a prod build

Sources/WhisperKit/Core/WhisperKit.swift
jkrukowski: previously `setupModels` would set `modelFolder` property. Right now we have to set either `modelFolder` or / and `mlxModelFolder` depending on the pipeline setup. In this context this function doesn't make sense any more

Makefile
jkrukowski: the rationale behind it is the this change will allow to simplify the `Package.swift` and the project structure. Models downloaded using `make` are needed basically just for testing, the `WhisperKit's` user will download models to some other location most likely (e.g. the default one provided by `swift-transformers`). Right now we have to provide a custom path and a list of files to exclude in the `Package.swift` just for the sake of having Models directory at the top level.

I'll revert it, this change is not needed for mlx integration

README.md
jkrukowski: Right now user has to pass either `model` and / or `mlxModel`, depending weather mlx is used or not

README.md
jkrukowski: reverted

ZachNagengast (approved):
Approved pending the setupModels change (also to run the tests)

ZachNagengast:
Test passing ✅

PR #190, opened 2024-08-01
Change `static var` stored properties to `static let`

fumoboy007:
As far as I can tell, these stored properties are not meant to be changed. Therefore, change them to be immutable. This change also makes these static properties concurrency-safe.


ZachNagengast:
Interesting, did you have a specifc use case where this was helping with thread safety for you? Or just a general improvement?

fumoboy007:
These properties won’t be usable in the stricter Swift 6 compiler without this change:
> error: reference to static property '…' is not concurrency-safe because it involves shared mutable state

ZachNagengast (approved):
Thanks for the context, looks ok to merge.

PR #192, opened 2024-08-06
Make additional initializers, functions, members public for extensibility

bpkeene:
* Allows use of default internal functions & member accesses which have increased protections when imported

* Initializers were Xcode generated: right click class name -> refactor -> generate memberwise initializers
   * memberwise initializer defaults to internal, mark as public.


ZachNagengast (approved):
Looks good 👍

PR #199, opened 2024-09-05
Add VoiceActivityDetector base class

a2they:
Add base class to allow different VAD implementations


ZachNagengast (approved):
Looks great!

PR #216, opened 2024-10-06
Add model support config fetching from model repo

ZachNagengast:
This PR adds capability to fetch supported models for a specific device based on the [config.json](https://huggingface.co/argmaxinc/whisperkit-coreml/blob/main/config.json) file in the model repo on HF. Config will update over time as new models and devices come out. Need for this arose with the latest whisper turbo model release.

Changes:
- Adds asynchronous `WhisperKit.recommendedRemoteModels()` to fetch the config.json file and return a `ModelSupport` object
- ~New optional parameters for synchronous `WhisperKit.recommendedModels()` function to try to fetch the remote config, and a timeout for the request~ This proved difficult to conform to swift concurrency as a static function for macOS 14.6 and below, so removed for now.
- Adds `Constants.fallbackModelSupportConfig` for when remote config is unavailable in offline environments
- Various supporting functions and models
- Updates `WhisperKit.deviceName()` to use a similar device identifier for mac as it does for iOS etc

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -60,7 +60,7 @@ jobs:
         run: make download-model MODEL=tiny
       - name: Install and discover destinations
         run: |
-          xcodebuild -downloadAllPlatforms
+          xcodebuild -downloadPlatorm ${{ matrix.run-config['name'] }}
           echo "Destinations for testing:"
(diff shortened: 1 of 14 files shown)


a2they (approved):
Awesome! 🚀

PR #236, opened 2024-10-31
Add convenience methods to detect if a model is installed

bmurray:
This adds convenience functions to detect if a model is installed locally, without querying the server.

### DIFF
--- a/Sources/WhisperKit/Core/WhisperKit.swift
+++ b/Sources/WhisperKit/Core/WhisperKit.swift
@@ -266,6 +266,79 @@ open class WhisperKit {
             throw error
         }
     }
+    public static func modelLocation(
+        variant: String,
+        downloadBase: URL? = nil,
+        from repo: String = "argmaxinc/whisperkit-coreml",
+        specificOnly: Bool = false
+    ) throws -> URL {
+        let saveRoot: URL
+        if let downloadBase {
+            saveRoot = downloadBase
+        } else {
+            let documents = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first!
+            saveRoot = documents.appending(component: "huggingface")
+        }
+        let modelSearchPath = "*\(variant.description)"
+        let repoDestination = saveRoot.appending(component: "models").appending(component: repo)
+
+        do {
+            Logging.debug(repoDestination.absoluteString)
+            let isInstalled = try? repoDestination.resourceValues(forKeys: [.isDirectoryKey]).isDirectory
+            guard let isInstalled, isInstalled else {
+                throw WhisperError.modelsUnavailable("Repo destination does not exist \(repoDestination.absoluteString)")
+            }
+            let dirFiles = try FileManager.default.contentsOfDirectory(atPath: repoDestination.path(percentEncoded: false))
+            let modelFiles = dirFiles.matching(glob: modelSearchPath)
+            var uniquePaths = Set(modelFiles.map { $0.components(separatedBy: "/").first! })
+            
+            var variantPath: String? = nil
+            
+            if uniquePaths.count == 1 {
+                variantPath = uniquePaths.first
+            } else if specificOnly {
+                // We only want the one specific model, and won't accept fuzzy fallbacks
+                throw WhisperError.modelsUnavailable("Multiple models found matching \"\(modelSearchPath)\" and specificOnly was set")
+            } else {
+                // If the model name search returns more than one unique model folder, then prepend the default "openai" prefix from whisperkittools to disambiguate
+                Logging.debug("Multiple models found matching \"\(modelSearchPath)\"")
+                let adjustedModelSearchPath = "*openai*\(variant.description)"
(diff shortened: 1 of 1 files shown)


ZachNagengast:
Thanks for adding this @bmurray looks useful! Can you think of a test to add to verify it is working? Can be part of the unit tests here: https://github.com/argmaxinc/WhisperKit/blob/main/Tests/WhisperKitTests/UnitTests.swift

Also, was the need for this derived from needing offline access to the models? Or some other motivation? We definitely want to make sure offline access is always available so if there was a regression there from a recent release we should fix that.

bmurray:
Yea I'm working on replacing WhisperCPP in my app with this, and this is a feature that I needed. It's more or less just detecting if the model is already installed or not, so I can provide feedback to users. I'm probably going to add some other additions here in a bit that provide other feedback, such as if the model is downloading. As far as I can tell, theres no visibility into whats happening if it "auto" downloads a model when a transcription starts. 

I'll take a look to see if theres a unit test that makes sense.

bmurray:
I took a look and I'm not sure where to put any tests for this that don't require downloading models just to test if they're installed. 

I updated the pull to do offline support. With download: false set, and the model already downloaded (either manually or automatically), the transcriber can run. It probably needs some more testing as theres a bunch of things that silently query websites.

ZachNagengast:
Ok cool, I think a good test would start in a similar way to [this one](https://github.com/argmaxinc/WhisperKit/blob/a9b92c4399e8a265b0bb48ff07d28c3c6d221790/Tests/WhisperKitTests/UnitTests.swift#L28-L33), where it just loads the model (which would imply its downloaded), then simply make sure the new functions correctly detect that it is installed.

Also I'd recommend keeping this interface the same `modelRepo: String?,` since folks will have a breaking change when they update to the new version.

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -266,12 +266,87 @@ open class WhisperKit {
             throw error
         }
     }
+    public static func modelLocation(
```
ZachNagengast, line 269: Should add a new line above this for code style. Will want to add docc for all of the functions in this file eventually but not a blocker.

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -266,12 +266,87 @@ open class WhisperKit {
...
+            }
+            
+            let truePath = repoDestination.appending(path: variantPath)
+            return truePath
+        } catch {
+            Logging.debug(error)
+            throw error
+        }
+    }
+    public static func modelInstalled(
```
ZachNagengast, line 329: New line above this as well

Tests/WhisperKitTests/UnitTests.swift
```diff
@@ -31,6 +31,46 @@ final class UnitTests: XCTestCase {
...
+        XCTAssertFalse(
+            WhisperKit.modelInstalled(variant: "THIS_MODEL_DOES_NOT_EXIST"),
+            "Model does not exist, but returned true"
+        )
+        XCTAssertTrue(
+            WhisperKit.modelInstalled(variant: "tiny"),
+            "Model was not installed"
+            )
+        XCTAssertTrue(
+            WhisperKit.modelInstalled(variant: "tiny.en"),
```
ZachNagengast, line 48: You'll need some method to download this one ahead of time before checking it is installed. https://github.com/argmaxinc/WhisperKit/blob/dd2eb733e9d151379092d47886a69cc07ac6d9f0/.github/workflows/unit-tests.yml#L60 you can do that by either adding a line `make download-model MODEL=tiny.en` here to download it, or just removing this tiny.en check and only running this test for tiny.

Tests/WhisperKitTests/UnitTests.swift
```diff
@@ -31,6 +31,46 @@ final class UnitTests: XCTestCase {
...
+            "Model was not installed"
+            )
+        let location = try WhisperKit.modelLocation(variant: "tiny")
+        let location2 = try WhisperKit.modelLocation(variant: "openai_whisper-tiny")
+        XCTAssertEqual(location, location2, "Auto fallback to OpenAI model returned a different result")
+    }
+    func testOfflineMode() async throws {
```
ZachNagengast, line 55: New line above this one

Tests/WhisperKitTests/UnitTests.swift
```diff
@@ -31,6 +31,46 @@ final class UnitTests: XCTestCase {
...
+    
+        let pipe = try await WhisperKit(WhisperKitConfig(model: "tiny.en", download: false))
+
+            let cancellable: AnyCancellable? = pipe.progress.publisher(for: \.fractionCompleted)
+                .removeDuplicates()
+                .withPrevious()
+                .sink { previous, current in
+                    if let previous {
+                        XCTAssertLessThan(previous, current)
```
ZachNagengast, line 64: Cool! 👍

ZachNagengast (approved):
Approving pending the review comments so that the tests will run.

PR #262, opened 2024-11-18
Add ability to prevent config.json being written to `~/Documents/huggingface/...`

iandundas:
This PR makes it possible to prevent this file being written during `WhisperKit.fetchModelSupportConfig()`, by allowing a `downloadBase: URL?` parameter to be passed. So it downloads to there instead.


a2they (approved):
Nice! It would be nice to pass downloadBase from`recommendedRemoteModels` as well.

ZachNagengast (approved):
Looks great to me 👍

PR #269, opened 2024-12-04
Typo in Model Descriptions

rk-helper:
It should be "Downloaded" for .downloaded, not "Downloading"


ZachNagengast (approved):
Nice catch 👍

PR #278, opened 2024-12-16
Audio: Fix taking a suffix of negative length from a collection

mattisssa:
# Overview
This pull request introduces critical bug fix to `AudioProcessor`, focusing on improving robustness and error handling in voice detection mechanisms

### Problem
In the original code, when calculating the number of energy values to consider for voice detection, there was a risk of creating a negative array index. 

This could happen if:
	•	The nextBufferInSeconds was very small
	•	The calculation resulted in a negative number

This would cause a runtime crash with the error: 
```Fatal error: Can't take a suffix of negative length from a collection```

CleanShot 2024-12-16 at 20 07 46@2x

### Solution
Prevent negative array indices


ZachNagengast (approved):
Looks good to me, thank you!

bimehasia25815 (commented):
A

PR #289, opened 2025-01-02
Updates the kvcache func in TextDecoder to work with Swift6

jdowd7:
Refactors this func to address the warning: 

> Capture of 'keyTensorPointer' with non-sendable type 'UnsafeMutableRawBufferPointer' in a @Sendable closure; this is an error in the Swift 6 language mode

- Simplified the index calculations by treating the 4D tensors as 2D (hidden_dim × seq_length)
- Added bounds checking to prevent out-of-range access
- Used direct array indexing with properly calculated linear indices
- Simplified memory copying using direct pointer access for the slices

Tested in our application and its working as expected

### DIFF
--- a/Sources/WhisperKit/Core/TextDecoder.swift
+++ b/Sources/WhisperKit/Core/TextDecoder.swift
@@ -308,45 +308,49 @@ public extension TextDecoding {
         return kvCache
     }
 
-    static func updateKVCache(keyTensor: MLMultiArray, keySlice: MLMultiArray,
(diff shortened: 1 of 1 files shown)


ZachNagengast:
Thanks @jdowd7! Do you have any metrics on if there is a speed difference between the two implementations?

jdowd7:
> Thanks @jdowd7! Do you have any metrics on if there is a speed difference between the two implementations?

Not yet, will benchmark and report back when I get a chance

PR #305, opened 2025-02-21
Improve SegmentSeeker word alignment

ZachNagengast:
These changes primarily focus on enhancing the `SegmentSeeker` class to fix some edge cases and make it more testable. It also adds public initializers to various structs and uses more flexible dependency versioning to allow usage of newer versions of swift-transformers if desired.


Sources/WhisperKit/Core/Models.swift
```diff
@@ -605,6 +657,39 @@ public struct TranscriptionSegment: Hashable, Codable {
...
     public var noSpeechProb: Float = 0.0
     public var words: [WordTiming]? = nil
+
+    /// Computed property for the duration of the segment
+    public var duration: Float {
+        return end - start
+    }
+
+    public init(
+        id: Int = 0,
```
a2they, line 667: since we set defaults here, we should remove redundancy above

Sources/WhisperKit/Core/Text/SegmentSeeker.swift
```diff
@@ -465,35 +469,73 @@ open class SegmentSeeker: SegmentSeeking {
...
+            lastSpeechTimestamp: lastSpeechTimestamp,
+            constrainedMedianDuration: constrainedMedianDuration,
+            maxDuration: maxDuration,
+            tokenizer: tokenizer
+        )
+
+        return updatedSegments
+    }
+
+    public func calculateWordDurationConstraints(alignment: [WordTiming]) -> (Float, Float) {
```
a2they, line 494: let's add names to Floats in the return tuple, i.e. `(median: Float, max: Float)`

Sources/WhisperKit/Core/Text/SegmentSeeker.swift
```diff
@@ -507,10 +549,42 @@ open class SegmentSeeker: SegmentSeeking {
...
+                // Retokenize the word if some tokens were filtered out
+                let word = timingTokens.count < timing.tokens.count ?
+                    tokenizer.decode(tokens: timingTokens) :
+                    timing.word
+
+                var start = (timeOffset + timing.start).rounded(2)
                 let end = (timeOffset + timing.end).rounded(2)
+
+                // Handle short duration words by moving their start time back if there's space
+                if end - start < constrainedMedianDuration / 4 {
```
a2they, line 561: nit: it would be nice to add example in comments for each of these if/else cases

a2they (approved):
lgtm

Issue #313, opened 2025-03-13
support for multi channel audio


kanishkave:
Hey does whisperkit support multichannel audio specially .caf apple files?

ZachNagengast:
This is in progress! If you have any specific audio files to test with or preferred algorithms for combining channels please let us know here.

Issue #134, opened 2024-05-04
Audio input captures only the first channel


iGerman00:
Great library and equally great demo app! I tried using the demo, and it works quite well with the built-in mic of my MacBook, but it seems like currently the microphone input only captures the first channel. This is fine in 90% of use cases, but my audio interface presents itself as a stereo input to the system, with the left channel being one physical input, and the right being another, and my mic is plugged into the right (so floatChannelData[1] instead of [0]), rendering the app unusable because there is no way to switch/combine those inputs using macOS or the app.

I suggest summing all inputs of the device, or at least give the user the option to choose what input channel to use.
Seems like this issue lies in [this line](https://github.com/argmaxinc/WhisperKit/blob/d6f50da1dc0b69fc53bb7d4a14f3972cc1353174/Sources/WhisperKit/Core/AudioProcessor.swift#L324), so it's outside of the actual demo app and in the library itself, though I'm not certain.

Would really appreciate a quick fix and push to TestFlight, I'm not trying to set up Xcode and compile the app myself at the moment - don't even have the disk space for that :P

ZachNagengast:
Thanks for raising this, I've heard some similar issues with input channels. I'm aware of some ways to combine the channels from stereo to mono with a mixer node, but will need to investigate. In the meantime the simplest solution would be to remap your mic input with an app like blackhole or soundflower - the testflight app was purely intended as an example for how to use the library, though I'm glad to hear some folks are getting use out of it!

amirvenus:
> Great library and equally great demo app! I tried using the demo, and it works quite well with the built-in mic of my MacBook, but it seems like currently the microphone input only captures the first channel. This is fine in 90% of use cases, but my audio interface presents itself as a stereo input to the system, with the left channel being one physical input, and the right being another, and my mic is plugged into the right (so floatChannelData[1] instead of [0]), rendering the app unusable because there is no way to switch/combine those inputs using macOS or the app.
> 
> I suggest summing all inputs of the device, or at least give the user the option to choose what input channel to use. Seems like this issue lies in [this line](https://github.com/argmaxinc/WhisperKit/blob/d6f50da1dc0b69fc53bb7d4a14f3972cc1353174/Sources/WhisperKit/Core/AudioProcessor.swift#L324), so it's outside of the actual demo app and in the library itself, though I'm not certain.
> 
[...]

PR #320, opened 2025-03-28
Multi channel audio merging

flashno:
Previous versions of WhisperKit was only using the first channel of Audio, regardless of whether it contained 2 or more channels. This update allows developers to combine all channels available, or choose specific channels to combine to feed into the transcription. To set the configuration on load of WhisperKit, you can pass it through to WhisperKitConfig similar to this:
```swift
        let config = WhisperKitConfig(
            ...
            audioInputConfig: AudioInputConfig(channelMode: .sumChannels([1, 3, 5]))
        )
```
This will sum the 2nd, 4th, and 6th channel (the first channel is index 0). 
Another way to pass the channelMode is to directly feed it into the loadAudio function like this:
```swift
        let audioArray = try AudioProcessor.loadAudioAsFloatArray(
            fromPath: audioPath,
            channelMode: .specificChannel(0)
        )
```
This option has been added to all audio processing API's that would read audio. 

⚠️ Important: This PR changes the default behavior to now sum and normalize all channels if the audio is not mono to begin with. 
The audio merging algorithm works like this:
We find the peak across all channels, check if the peak of the mono (summed) version is higher than any of the peaks of the channels, then we multiply the whole track so that the peak of the mono channel matches the peak of the loudest channel.

Eg: Top mono (merged) buffer, bottom individual channels (pre-merge)
image

Here you can see how the merged audio maintains the same loudness as the original multi-channel audio file, and you can see the total merged waveform of all the channels. 

Resolves these two issues:
https://github.com/argmaxinc/WhisperKit/issues/134
https://github.com/argmaxinc/WhisperKit/issues/313


ZachNagengast (approved):
Looks good 👍

PR #324, opened 2025-04-15
Use remote models for test

ZachNagengast:
This will allow the unit tests to be run from the example WhisperAX app without needing to run `make setup` or download any models manually, which is useful for testing environments that use physical devices and require a host app.


Tests/WhisperKitTests/TestUtils.swift
```diff
@@ -142,49 +147,46 @@ extension XCTestCase {
...
+        let modelName: String
         switch variant {
             case .largev3:
-                modelPath = try largev3ModelPath()
+                modelName = "large-v3"
             default:
-                modelPath = try tinyModelPath()
+                modelName = "tiny"
         }
         let computeOptions = ModelComputeOptions(
```
a2they, line 157: why cpu only? should this be only if target is simulator?

a2they (approved):

Tests/WhisperKitTests/TestUtils.swift
```diff
@@ -142,49 +147,46 @@ extension XCTestCase {
         file: StaticString = #file,
         line: UInt = #line
     ) async throws -> [TranscriptionResult] {
-        let modelPath: String
+        let modelName: String
         switch variant {
             case .largev3:
-                modelPath = try largev3ModelPath()
+                modelName = "large-v3"
```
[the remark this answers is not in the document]
ZachNagengast: This code path isnt actively tested in the unit tests, and has been superseded but the regression test suite, so probably just a todo to remove it in the future

Tests/WhisperKitTests/TestUtils.swift
ZachNagengast: Pretty old commit, may have been intended to keep the unit test output consistent across chips.

PR #326, opened 2025-04-22
Fix modelSupport where prefix overlaps with different hardware chips

a2they:
Before this change, the code was matching first prefix which could lead to selecting incorrect configuration. For example, iPad 13,1 (A14) is prefix for iPad13,10, iPad13,11, iPad13,16, and iPad13,17 (M2).

Fixing this issue by using longest prefix match.

Tested with a unit test that covers all iPad / iPhone devices.


ZachNagengast (approved):
Looks good 👍

PR #333, opened 2025-06-03
fix: the issue where filenames containing dots are not handled correctly when  generating the report path

JimLiu:
**Reproduce steps:**

transcribe a filename contains `.`
 `swift run whisperkit-cli transcribe --audio-path "~/Downloads/1.video.mp4" --verbose --word-timestamps --report --report-path ~/Downloads`

**Expected result:** `~/Downloads/1.video.json`

**actual result:** `~/Downloads/1.json`


ZachNagengast (approved):
Nice fix! Thank you for contributing

PR #347, opened 2025-08-25
Make EnergyVAD public

finnvoor:
Allows using `EnergyVAD` for standalone VAD similar to `VoiceActivityDetector.modelVAD()`

### DIFF
--- a/Sources/WhisperKit/Core/Audio/EnergyVAD.swift
+++ b/Sources/WhisperKit/Core/Audio/EnergyVAD.swift
@@ -5,8 +5,8 @@ import Foundation
 
 /// Voice activity detection based on energy threshold
 @available(macOS 13, iOS 16, watchOS 10, visionOS 1, *)
-final class EnergyVAD: VoiceActivityDetector {
-    var energyThreshold: Float
+public final class EnergyVAD: VoiceActivityDetector {
+    public let energyThreshold: Float
 
     /// Initialize a new EnergyVAD instance
     /// - Parameters:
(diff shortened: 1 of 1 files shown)


Sources/WhisperKit/Core/Audio/EnergyVAD.swift
```diff
@@ -5,16 +5,16 @@ import Foundation
 
 /// Voice activity detection based on energy threshold
 @available(macOS 13, iOS 16, watchOS 10, visionOS 1, *)
-final class EnergyVAD: VoiceActivityDetector {
-    var energyThreshold: Float
+public final class EnergyVAD: VoiceActivityDetector {
+    public let energyThreshold: Float
```
ZachNagengast, line 9: Can you think of any use case to change the threshold on the vad outside of the init method? Fine if not, just noticing the change from var to let.

finnvoor: No particular reason I switched to `let`, I can't imagine much use for changing it after but happy to change it back or keep it as is

ZachNagengast: Ok, this is mainly used for VAD chunking so probably not needed, but it would add extra flexibility with var if for some reason it needs to be updated outside of the init. Its a small class though so easy to reinitialize. Thoughts?

ZachNagengast:
@finnvoor The CI was recently fixed in main - if you want to merge that in, tests should pass

PR #360, opened 2025-09-23
Set watchOS and visionOS minimum deployment targets in Package manifest instead

naykutguven:
This pull request removes the `@available(macOS 13, iOS 16, watchOS 10, visionOS 1, *)` availability annotations across the codebase since the minimum deployment targets are already set as macOS 13 and iOS 16 in `Package.swift`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -9,6 +9,8 @@ let package = Package(
     platforms: [
         .iOS(.v16),
         .macOS(.v13),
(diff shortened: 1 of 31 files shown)


naykutguven:
Hey @ZachNagengast wdyt about this one?

naykutguven:
> Makes sense, nice 👍

awesome, thank you! ready when you are 👍

PR #367, opened 2025-10-16
Prepare for new swift-transformers, add TranscribeTask hooks

ZachNagengast:
This PR includes some extensions from swift-transformers > 1.0.0 that were being used and are now private, which will help when upgrading in the near future. 

It also makes `TranscribeTask` overrideable and includes a couple new functions that may be useful for running logic on a single window alongside the transcribe process.
⚠️ These hooks will await for the logic to complete if not wrapped in a detached task, which could block the transcription process, so use with caution - intended for advanced use cases.

Additionally, minor changes to improve the test CI further.

### DIFF
--- a/Sources/WhisperKit/Utilities/Extensions+Internal.swift
+++ b/Sources/WhisperKit/Utilities/Extensions+Internal.swift
@@ -4,6 +4,24 @@
 import AVFoundation
 import CoreML
 
+extension MLMultiArray {
+    /// All values will be stored in the last dimension of the MLMultiArray (default is dims=1)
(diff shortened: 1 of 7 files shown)


Sources/WhisperKit/Utilities/Extensions+Internal.swift
```diff
@@ -35,6 +53,26 @@ extension Array where Element: Hashable {
     }
 }
 
+extension String {
```
pcuenca, line 56: Ah, didn't know you were using these, sorry!

PR #386, opened 2025-11-25
Pin Xcode version as 26 for Github workflows

naykutguven:
This pull request updates the iOS, watchOS, visionOS, and Xcode versions used in the GitHub Actions workflows for development and release testing. The primary goal is to pin all relevant jobs to version 26.0 for consistency and reproducibility, instead of using the latest stable or 26.1 versions.

Workflow version pinning:

* Updated `ios-version`, `watchos-version`, `visionos-version`, and `xcode-version` from "26.1" or "latest-stable" to "26.0" in `.github/workflows/development-tests.yml` and `.github/workflows/release-tests.yml` to ensure all CI jobs use the same, predictable environment.
* Changed the default `xcode-version` input in `.github/workflows/unit-tests.yml` from "latest-stable" to "26.0" for more deterministic builds.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -58,7 +58,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: maxim-lobanov/setup-xcode@v1
         with:
-          xcode-version: ${{ inputs.xcode-version || '26.0.1' }}
+          xcode-version: ${{ inputs.xcode-version || '26.0' }}
       - name: Setup environment
         run: make setup
       - name: Setup Cache


.github/workflows/unit-tests.yml
```diff
@@ -58,7 +58,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: maxim-lobanov/setup-xcode@v1
         with:
-          xcode-version: ${{ inputs.xcode-version || 'latest-stable' }}
+          xcode-version: ${{ inputs.xcode-version || '26.0.1' }}
```
ZachNagengast: I think this was mixed up with ios version?
```suggestion
          xcode-version: ${{ inputs.xcode-version || '26.0' }}
```

naykutguven: It was on purpose to be more specific with it but I'll try 26.0 👍

PR #394, opened 2026-01-08
Using the endpoint parameter in Whisperkit and downloading CoreML and tokenizer

lemo366:
Using the endpoint parameter in Whisperkit and downloading CoreML and tokenizer from hf-mirror.com.


ZachNagengast:
@lemo366 Thanks for taking this on, but looks like your changes didnt get pushed? This PR seems to be missing what is indicated in your description.

PR #396, opened 2026-01-10
Using the endpoint parameter in Whisperkit and downloading CoreML and tokenizer

lemo366:
Allows users to specify the `endpoint` via parameters when initializing WhisperKit, thus enabling the download of CoreML models and tokenizers from domestic mirror sites (such as hf-mirror.com).
```swift
let endpoints = "https://hf-mirror.com"
let whisperkit = try await WhisperKit(
      model: modelName,
      verbose: true,
      logLevel: .info, 
      prewarm: true,
      load: true,
      download: true,
      endpoint: endpoint
)
```

### DIFF
--- a/Sources/WhisperKit/Utilities/ModelUtilities.swift
+++ b/Sources/WhisperKit/Utilities/ModelUtilities.swift
@@ -22,7 +22,7 @@ public struct ModelUtilities {
         tokenizerFolder: URL? = nil,
         additionalSearchPaths: [URL] = [],
         useBackgroundSession: Bool = false,
-        endpoint: String
+        endpoint: String = Constants.defaultRemoteEndpoint
     ) async throws -> WhisperTokenizer {
         let tokenizerName = tokenizerNameForVariant(pretrained)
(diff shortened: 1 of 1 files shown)


ZachNagengast:
Thanks @lemo366 this looks great! Do the unit tests pass with hf-mirror as the default? Might be worth including at least one test that uses hf-mirror as the endpoint for a simple download.

Sources/WhisperKit/Utilities/ModelUtilities.swift
```diff
@@ -21,7 +21,8 @@ public struct ModelUtilities {
         for pretrained: ModelVariant,
         tokenizerFolder: URL? = nil,
         additionalSearchPaths: [URL] = [],
-        useBackgroundSession: Bool = false
+        useBackgroundSession: Bool = false,
+        endpoint: String
```
ZachNagengast: This should allow you to undo the changes in UnitTests by providing a default value for endpoint
```suggestion
        endpoint: String = Constants.defaultRemoteEndpoint
```

PR #401, opened 2026-01-18
Mute-style input suppression without pausing AVAudioEngine

shura-v:
### Context
Some apps need to temporarily suppress input without pausing AVAudioEngine, since `pause()` can cause resume issues in certain iOS scenarios.

### Changes
Introduced simple mute-style switch (`isInputSuppressed`, `setInputSuppressed`) to AudioProcessor: when suppressed, we pass through silence so timing remains consistent.

Mute is implemented allocation-free by zeroing the samples in place (via Accelerate, which is already imported), to keep the tap path leaner.

### DIFF
--- a/Sources/WhisperKit/Core/Audio/AudioProcessor.swift
+++ b/Sources/WhisperKit/Core/Audio/AudioProcessor.swift
@@ -1011,12 +1011,7 @@ public extension AudioProcessor {
             }
 
             var newBufferArray = Self.convertBufferToArray(buffer: buffer)
-            if self.isInputSuppressed {
-                newBufferArray.withUnsafeMutableBufferPointer { ptr in
-                    guard let base = ptr.baseAddress else { return }
-                    vDSP_vclr(base, 1, vDSP_Length(ptr.count))
-                }
-            }
+            self.suppressInputIfNeeded(&newBufferArray)
             self.processBuffer(newBufferArray)
         }
 
@@ -1096,5 +1091,13 @@ public extension AudioProcessor {
         // Stop the audio engine
         audioEngine?.stop()
         audioEngine = nil
+    }
+
+    func suppressInputIfNeeded(_ buffer: inout [Float]) {
+        guard isInputSuppressed else { return }
+        buffer.withUnsafeMutableBufferPointer { ptr in
+            guard let base = ptr.baseAddress else { return }
+            vDSP_vclr(base, 1, vDSP_Length(ptr.count))
+        }
     }
 }
--- a/Sources/WhisperKit/Core/Audio/AudioProcessor.swift
+++ b/Sources/WhisperKit/Core/Audio/AudioProcessor.swift
@@ -213,10 +213,11 @@ open class AudioProcessor: NSObject, AudioProcessing {
 
     public var audioBufferCallback: (([Float]) -> Void)?
     public var minBufferLength = Int(Double(WhisperKit.sampleRate) * 0.1) // 0.1 second of audio at 16,000 Hz
+    public private(set) var isInputSuppressed = false
 
-    /// Override to drop buffers without pausing the audio engine.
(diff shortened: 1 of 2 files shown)


Sources/WhisperKit/Core/Audio/AudioProcessor.swift
```diff
@@ -213,6 +213,11 @@ open class AudioProcessor: NSObject, AudioProcessing {
 
     public var audioBufferCallback: (([Float]) -> Void)?
     public var minBufferLength = Int(Double(WhisperKit.sampleRate) * 0.1) // 0.1 second of audio at 16,000 Hz
+
+    /// Override to drop buffers without pausing the audio engine.
+    open func shouldAcceptBuffer(_ buffer: AVAudioPCMBuffer) -> Bool {
```
naykutguven: Just out of curiosity, isn't this more of a responsibility a delegate should take on? True by default = also true if the delegate is nil. and the signature conveys a meaning that it should ask something else for permission

ZachNagengast: Fair question - @shura-v do you anticipate doing some processing on the buffer coming into the function? If so, it would make sense to have a "preprocessing" hook similar to this https://github.com/argmaxinc/WhisperKit/blob/ba094495e40ff255f30ca4602e2c75994f255df7/Sources/WhisperKit/Core/TranscribeTask.swift#L42-L46 which could return the processed buffer from this open function - but this will be called very often and could introduce lag into the pipeline. If it's purely to pause or allow the buffer to go into processing, it may be better to have a boolean property that can be an early return for the input processing itself, rather than an open function here, because it  would require subclassing by the consumers to implement. 

The other tricky thing with blocking audio entirely from getting processed is the timings of the transcription is generally based on how much audio has been passed into the pipeline - by sythetically blocking buffers from going through you'd want to make sure the word timings are still accurate. One final idea to suggest is perhaps instead of early return, you could call this a "mute" function that if set would swap out the audio being passed in for all 0 silent data. Curious your thoughts.

shura-v: Thanks for the review.

I didn't actually intend to use that argument for processing, it was mainly a convenience hook.

I reworked this into a simple mute-style switch as suggested (`isInputSuppressed`, `setInputSuppressed`): when suppressed, we pass through silence instead of dropping buffers, so timing remains consistent.

Mute is implemented allocation-free by zeroing the samples in place (via Accelerate, which is already imported), to keep the tap path leaner.

Note: resampling still happens when muted to keep the buffer length/timing consistent.

ZachNagengast (approved):
Cool! This makes sense to me 👍

Sources/WhisperKit/Core/Audio/AudioProcessor.swift
```diff
@@ -213,6 +213,12 @@ open class AudioProcessor: NSObject, AudioProcessing {
 
     public var audioBufferCallback: (([Float]) -> Void)?
     public var minBufferLength = Int(Double(WhisperKit.sampleRate) * 0.1) // 0.1 second of audio at 16,000 Hz
+    public private(set) var isInputSuppressed = false
+
+    /// Suppress input buffers by replacing them with silence while keeping timing intact.
+    public func setInputSuppressed(_ isSuppressed: Bool) {
+        isInputSuppressed = isSuppressed
+    }
```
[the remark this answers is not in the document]
shura-v: Done in [Add tests for input suppression](https://github.com/argmaxinc/WhisperKit/pull/401/commits/33f7ba2fdd71c31b9d9e2b1c0659ffd4718e944c), [Add toggle test for input suppression](https://github.com/argmaxinc/WhisperKit/pull/401/commits/3362a7498c44cf67bb56dad0ab1898a63febcfbd)

ZachNagengast:
Looks great, thanks for the follow up with tests 👍

PR #402, opened 2026-01-18
AudioProcessor: fix teardown to avoid StartIO/thread warnings on some Bluetooth devices

shura-v:
On some Bluetooth devices (e.g. Sony WH-1000XM5), restarting recording may fail to start audio capture.
**Fix:** Strengthen AVAudioEngine teardown to fully release the input node and avoid stale audio graph state.

### Changes

- Remove tap from `inputNode` first (not only `attachedNodes`)
- Disconnect input node (`disconnectNodeInput`)
- `reset()` engine before releasing

### DIFF
--- a/Sources/WhisperKit/Core/Audio/AudioProcessor.swift
+++ b/Sources/WhisperKit/Core/Audio/AudioProcessor.swift
@@ -1078,14 +1078,22 @@ public extension AudioProcessor {
     func stopRecording() {
         guard let engine = audioEngine else { return }
 
-        // Remove the tap from the input node first; other nodes are secondary.
+        // Remove tap from the input node explicitly.
         engine.inputNode.removeTap(onBus: 0)
+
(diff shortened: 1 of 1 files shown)


ZachNagengast (approved):
Nice fix! Did this resolve the issue for your Bluetooth device?

shura-v:
> Nice fix! Did this resolve the issue for your Bluetooth device?

Yes, it did. Works like a charm. Tested on Sony WH-1000XM5 (macOS 26). Restart recording reliably resumes audio capture, and the StartIO/thread warnings are gone.

a2they (approved):

ZachNagengast:
Looks great! Thanks for contributing this. There is another bug in this section that causes a crash if `stopRecording` is called from two different threads at the same time, but that can be resolved with Swift 6 concurrency or equivalent in another PR. Good to merge 👍

PR #404, opened 2026-01-21
Update Swift toolchain to 6.2 and build settings in Package manifest

naykutguven:
This pull request updates the Swift package configuration to adopt Swift 6.2 and enable upcoming concurrency features across all targets. The main changes involve updating the Swift tools version, introducing a shared set of concurrency-related Swift settings, and applying them consistently to all targets in the Package.swift file.

The language mode is still set to Swift 5 since the project doesn't fully compile in Swift 6 yet. Once the changes required for Swift 6 migration are complete, we can set it to Swift 6.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -4,6 +4,11 @@
 import PackageDescription
 import Foundation
 
+let approachableConcurrencySettings: [SwiftSetting] = [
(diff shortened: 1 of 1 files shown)


naykutguven:
@ZachNagengast could you re-trigger the failed jobs so we can merge this PR if everything is ok?

PR #407, opened 2026-01-25
Move `decodeWithFallback` outside the enclosing method

naykutguven:
This pull request refactors `decodeWithFallback` method and move it outside the enclosing function to fix concurrency errors. The logic remains the same.

### Decoding fallback logic

* Refactored the `decodeWithFallback` method to accept explicit references to `decoderInputs`, `detectedLanguage`, and `fallbackCount`, rather than relying on implicit state. This makes the fallback mechanism clearer and easier to maintain.
* Fixes the following Swift Concurrency errors:
  * `Sending task-isolated 'decoderInputs' to @concurrent local function 'decodeWithFallback(encoderSegment:decodingOptions:callback:)' risks causing data races between @concurrent and task-isolated uses`
  * `Sending task-isolated 'encoderOutput' to @concurrent local function 'decodeWithFallback(encoderSegment:decodingOptions:callback:)' risks causing data races between @concurrent and task-isolated uses`
  * `Sending task-isolated 'self' to @concurrent local function 'decodeWithFallback(encoderSegment:decodingOptions:callback:)' risks causing data races between @concurrent and task-isolated uses`

### DIFF
--- a/Sources/WhisperKit/Core/TranscribeTask.swift
+++ b/Sources/WhisperKit/Core/TranscribeTask.swift
@@ -177,7 +177,6 @@ open class TranscribeTask {
                     decodingOptions: options,
                     decoderInputs: &decoderInputs,
                     detectedLanguage: &detectedLanguage,
-                    fallbackCount: &fallbackCount,
                     prefilledCacheSize: prefilledCacheSize,
                     callback: decodingCallback
                 )
@@ -329,7 +328,6 @@ open class TranscribeTask {
         decodingOptions options: DecodingOptions,
         decoderInputs: inout any DecodingInputsType,
         detectedLanguage: inout String?,
-        fallbackCount: inout Int,
         prefilledCacheSize: Int,
         callback: TranscriptionCallback? = nil
     ) async throws -> DecodingResult {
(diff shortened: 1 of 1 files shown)


Sources/WhisperKit/Core/TranscribeTask.swift
```diff
@@ -172,7 +172,15 @@ open class TranscribeTask {
 
                 try Task.checkCancellation()
                 // Send to decoder to predict text tokens with fallback
-                let decodingResult = try await decodeWithFallback(encoderSegment: encoderOutput, decodingOptions: options, callback: decodingCallback)
+                let decodingResult = try await decodeWithFallback(
+                    encoderSegment: encoderOutput,
+                    decodingOptions: options,
+                    decoderInputs: &decoderInputs,
+                    detectedLanguage: &detectedLanguage,
+                    fallbackCount: &fallbackCount,
```
ZachNagengast: Do these need to be inout or can they just be returned from the function?

naykutguven: `decoderInputs` is shared between iterations and mutated before & after & in `decodeWithFallback`. `detectedLanguage` is determined by two different methods inside `decodeWithFallback` in each iteration. This one _can_ be returned from the function but it wouldn't look nice IMO considering the function name. I chose inout parameters to have the least impact on the current logic.

`fallbackCount`, however, can be just removed and kept within the function. doing it now

PR #421, opened 2026-02-16
fix: replace force-unwraps in TextDecoder with guard statements

### LINKED ISSUE #414: TextDecoder.decodeText force-unwraps cause crashes during repeated live transcription
## Summary

`TextDecoder.decodeText` contains force-unwraps that crash the app during repeated rapid transcription calls (e.g., live transcription preview that re-transcribes every ~200ms).

## Crash Stack

```
EXC_BREAKPOINT / SIGTRAP
_assertionFailure → TextDecoder.decodeText(from:using:sampler:options:callback:)
→ TranscribeTask.run → WhisperKit.transcribe → WhisperKit.transcribe(audioPath:)
```

## Root Cause

Two force-unwraps in `TextDecoder.decodeText` ([TextDecoder.swift](https://github.com/argmaxinc/WhisperKit/blob/main/Sources/WhisperKit/Core/TextDecoder.swift)):

1. **Line 747**: `decoderInputs.initialPrompt.last!` — crashes if `initialPrompt` is empty
2. **Line 829**: `decoderOutput.logits!` — crashes if logits is nil

These are fine for single-shot transcription but become problematic during sustained rapid calls where internal state can degrade.

## Reproduction

- Create two WhisperKit instances (main + live)
- Call `transcribe(audioPath:)` on the live instance in a tight loop (~5-10 calls/sec) during recording
- After extended use (minutes to hours), the decoder state accumulates corruption and one of the force-unwraps hits nil

## Environment

- WhisperKit v0.15.0
- macOS 26.2, Apple Silicon
- Model: openai_whisper-small
- Using separate WhisperKit instances for main and live transcription

## Suggested Fix

Replace the force-unwraps with guard statements that throw errors instead of crashing:

```swift
// Line 747
guard let firstToken = decoderInputs.initialPrompt.last else {
    throw WhisperError.decodingFailed("initialPrompt is empty")
}
var nextToken: Int = firstToken

// Line 829
guard let logits = decoderOutput.logits else {
    throw WhisperError.decodingLogitsFailed("Logits output is nil")
}
```

This would allow callers to handle the error gracefully instead of crashing the entire app.

## Related

I noticed #413 and #412 are addressing thread safety — this may be related to the same class of issues.

william-laverty:
## Summary

Replaces three force-unwraps in `TextDecoder.decodeText` that cause `EXC_BREAKPOINT` crashes during rapid repeated transcription calls (e.g., live transcription preview at ~200ms intervals).

## Changes

| Location | Before | After |
|----------|--------|-------|
| Line 747 | `decoderInputs.initialPrompt.last!` | `guard let` + `throw WhisperError.decodingFailed` |
| Line 683 | `decoderOutput.logits!` (language detection) | `guard let` + `throw WhisperError.decodingLogitsFailed` |
| Line 829 | `decoderOutput.logits!` (token decoding) | `guard let` + `throw WhisperError.decodingLogitsFailed` |

All three now throw descriptive `WhisperError` cases that already exist in the codebase, allowing callers to handle the error gracefully instead of crashing.

## Context

During sustained rapid transcription calls (5-10/sec for live preview), decoder state can degrade and hit these nil/empty conditions. The force-unwraps cause hard crashes instead of recoverable errors.

Fixes #414

### DIFF
--- a/Sources/WhisperKit/Core/TextDecoder.swift
+++ b/Sources/WhisperKit/Core/TextDecoder.swift
@@ -680,7 +680,10 @@ open class TextDecoder: TextDecoding, WhisperMLModel {
         // MARK: Non-inference
 
         // Update predicted token as current
-        let logits = languageLogitsFilter.filterLogits(decoderOutput.logits!, withTokens: currentTokens)
+        guard let decoderLogits = decoderOutput.logits else {
+            throw WhisperError.decodingLogitsFailed("Language detection failed: decoder output logits are nil")
+        }
+        let logits = languageLogitsFilter.filterLogits(decoderLogits, withTokens: currentTokens)
 
         // MARK: Sampling
 
@@ -744,7 +747,10 @@ open class TextDecoder: TextDecoding, WhisperMLModel {
         let prefilledIndex = decoderInputs.cacheLength[0].intValue
         let initialPromptIndex = decoderInputs.initialPrompt.count
         var currentTokens: [Int] = decoderInputs.initialPrompt
-        var nextToken: Int = decoderInputs.initialPrompt.last!
+        guard let lastPromptToken = decoderInputs.initialPrompt.last else {
+            throw WhisperError.decodingFailed("Initial prompt is empty, cannot begin decoding")
+        }
+        var nextToken: Int = lastPromptToken
         var logProbs: [Float] = Array(repeating: 0, count: currentTokens.count)
 
         // Logits filters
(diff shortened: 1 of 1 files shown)


ZachNagengast:
Little confused why this was opened, it looks like you already have a PR open here with the same diff? https://github.com/argmaxinc/WhisperKit/pull/417

william-laverty:
Apologies for the duplicate — this is the same change as #417. Closing this one.

PR #423, opened 2026-02-18
Implement an actor worker pool for multiple audio files transcription

naykutguven:
## Summary

This PR refactors multi-audio transcription to use an actor-based worker pool while keeping parallelism and fixing Swift concurrency boundaries in `WhisperKit`.

### What changed

- Replaced batch fan-out logic with a `WhisperKitTranscriptionWorker` actor pool in `Sources/WhisperKit/Core/WhisperKit.swift`.
- Workers run `TranscribeTask` instances built from shared dependencies.
- Added internal `transcribeWithOptions(..., parentProgress:)` path and preserved result ordering by index.
- Updated batch progress handling to parent/child progress composition for each audio item/chunk.
- Ensured models are loaded before worker execution (`loadModels()` in batch entry path).
- Added a safety fallback to sequential mode when worker creation returns an empty array (prevents modulo-by-zero crash).
- Tightened visibility of internal helpers where possible.

## Why

`WhisperKit` is non-Sendable, and previous task-group parallelization crossed concurrency boundaries in a way that could trigger sendability issues.  
This approach keeps concurrency isolated through actors, preserves throughput, and avoids sending `WhisperKit` across tasks.

## Risk / Tradeoffs

- `SharedTranscribeTaskDependencies` is marked `@unchecked Sendable`; this relies on component-level thread safety guarantees for pipeline implementations.
- Behavior should be equivalent, but progress reporting internals are now parent/child-based and were touched in multiple paths.

## Migration Notes

- No public API changes intended.

### DIFF
--- a/Sources/WhisperKit/Core/TranscriptionWorker.swift
+++ b/Sources/WhisperKit/Core/TranscriptionWorker.swift
@@ -5,7 +5,6 @@ import Foundation
 
 actor TranscriptionWorker {
     private let dependencies: TranscribeTaskDependencies
-    private var transcribeTask: TranscribeTask?
 
     init(dependencies: TranscribeTaskDependencies) {
         self.dependencies = dependencies
@@ -120,7 +119,6 @@ actor TranscriptionWorker {
         progress: Progress
     ) async throws -> [TranscriptionResult] {
         let workerTask = dependencies.makeTranscribeTask(progress: progress)
-        transcribeTask = workerTask
         workerTask.segmentDiscoveryCallback = segmentCallback
 
         let transcribeTaskResult = try await workerTask.run(
--- a/Sources/WhisperKit/Core/WhisperKit.swift
+++ b/Sources/WhisperKit/Core/WhisperKit.swift
@@ -795,94 +795,67 @@ open class WhisperKit {
         }
         let batchProgress = parentProgress ?? Progress(totalUnitCount: Int64(audioArrays.count))
         batchProgress.totalUnitCount = max(batchProgress.totalUnitCount, Int64(audioArrays.count))
-        defer {
-            batchProgress.completedUnitCount = batchProgress.totalUnitCount
-        }
         
         if requestedWorkerCount > 1 {
             let workers = createSharedTranscriptionWorkers(count: requestedWorkerCount)
-            guard !workers.isEmpty else {
-                return await transcribeSequentially(
-                    audioArrays: audioArrays,
-                    decodeOptionsArray: decodeOptionsArray,
-                    seekOffsets: seekOffsets,
-                    callback: callback,
-                    batchProgress: batchProgress
-                )
-            }
+            if !workers.isEmpty {
+                let defaultSegmentDiscoveryCallback = self.segmentDiscoveryCallback
 
-            let defaultSegmentDiscoveryCallback = self.segmentDiscoveryCallback
-
-            let indexedResults = await withTaskGroup(of: (index: Int, result: Result<[TranscriptionResult], Swift.Error>).self) { taskGroup in
-                for (audioIndex, audioArray) in audioArrays.enumerated() {
-                    let childProgress = Progress(totalUnitCount: 1)
-                    batchProgress.addChild(childProgress, withPendingUnitCount: 1)
-                    
-                    // Setup callback to keep track of batches and chunks
-                    let indexedAudioCallback: TranscriptionCallback = { progress in
-                        var indexedProgress = progress
-                        indexedProgress.windowId = audioIndex
-                        return callback?(indexedProgress)
-                    }
-
-                    // Setup segment callback to track chunk seek positions for segment discovery
-                    let indexedSegmentCallback: SegmentDiscoveryCallback? = if let seekOffsets {
-                        { [segmentDiscoveryCallback] segments in
-                            let seekOffset = seekOffsets[audioIndex]
-                            var adjustedSegments = segments
-                            for i in 0..<adjustedSegments.count {
-                                adjustedSegments[i].seek += seekOffset
-                            }
-                            segmentDiscoveryCallback?(adjustedSegments)
+                let indexedResults = await withTaskGroup(of: (index: Int, result: Result<[TranscriptionResult], Swift.Error>).self) { taskGroup in
+                    for (audioIndex, audioArray) in audioArrays.enumerated() {
+                        let childProgress = Progress(totalUnitCount: 1)
+                        batchProgress.addChild(childProgress, withPendingUnitCount: 1)
+                        
+                        // Setup callback to keep track of batches and chunks
+                        let indexedAudioCallback: TranscriptionCallback = { progress in
+                            var indexedProgress = progress
+                            indexedProgress.windowId = audioIndex
+                            return callback?(indexedProgress)
                         }
-                    } else {
-                        defaultSegmentDiscoveryCallback
+
+                        // Setup segment callback to track chunk seek positions for segment discovery
+                        let indexedSegmentCallback: SegmentDiscoveryCallback? = if let seekOffsets {
+                            { [segmentDiscoveryCallback] segments in
+                                let seekOffset = seekOffsets[audioIndex]
+                                var adjustedSegments = segments
+                                for i in 0..<adjustedSegments.count {
+                                    adjustedSegments[i].seek += seekOffset
+                                }
+                                segmentDiscoveryCallback?(adjustedSegments)
+                            }
+                        } else {
+                            defaultSegmentDiscoveryCallback
+                        }
+
+                        let worker = workers[audioIndex % workers.count]
+                        let decodeOptions = decodeOptionsArray[audioIndex]
+                        
+                        taskGroup.addTask {
+                            await worker.transcribe(
+                                index: audioIndex,
+                                audioArray: audioArray,
+                                decodeOptions: decodeOptions,
+                                callback: indexedAudioCallback,
+                                segmentCallback: indexedSegmentCallback,
+                                progress: childProgress
+                            )
+                        }
                     }
 
-                    let worker = workers[audioIndex % workers.count]
-                    let decodeOptions = decodeOptionsArray[audioIndex]
-                    
-                    taskGroup.addTask {
-                        await worker.transcribe(
-                            index: audioIndex,
-                            audioArray: audioArray,
-                            decodeOptions: decodeOptions,
-                            callback: indexedAudioCallback,
-                            segmentCallback: indexedSegmentCallback,
-                            progress: childProgress
-                        )
+                    var batchResult = [(index: Int, result: Result<[TranscriptionResult], Swift.Error>)]()
+                    batchResult.reserveCapacity(audioArrays.count)
+                    for await result in taskGroup {
+                        batchResult.append(result)
                     }
+                    return batchResult
                 }
 
-                var batchResult = [(index: Int, result: Result<[TranscriptionResult], Swift.Error>)]()
-                batchResult.reserveCapacity(audioArrays.count)
-                for await result in taskGroup {
-                    batchResult.append(result)
-                }
-                return batchResult
+                return indexedResults
+                    .sorted(by: { $0.index < $1.index })
+                    .map { $0.result }
             }
-
-            return indexedResults
-                .sorted(by: { $0.index < $1.index })
-                .map { $0.result }
         }
         
-        return await transcribeSequentially(
-            audioArrays: audioArrays,
-            decodeOptionsArray: decodeOptionsArray,
-            seekOffsets: seekOffsets,
-            callback: callback,
-            batchProgress: batchProgress
-        )
-    }
-    
-    private func transcribeSequentially(
-        audioArrays: [[Float]],
-        decodeOptionsArray: [DecodingOptions?],
-        seekOffsets: [Int]?,
-        callback: TranscriptionCallback?,
-        batchProgress: Progress
-    ) async -> [Result<[TranscriptionResult], Swift.Error>] {
         let defaultSegmentDiscoveryCallback = self.segmentDiscoveryCallback
         var result = [Result<[TranscriptionResult], Swift.Error>]()
         result.reserveCapacity(audioArrays.count)
@@ -933,9 +906,11 @@ open class WhisperKit {
     
     private func createSharedTranscriptionWorkers(count: Int) -> [TranscriptionWorker] {
         guard count > 0 else {
+            Logging.info("Worker pool creation skipped because requested worker count is \(count); falling back to sequential transcription for this batch.")
             return []
         }
         guard let tokenizer else {
+            Logging.info("Tokenizer unavailable for worker pool creation; falling back to sequential transcription for this batch.")
             return []
         }
         let dependencies = TranscribeTaskDependencies(


Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -1039,19 +1195,197 @@ open class WhisperKit {
...
+            if resetSharedProgressOnCompletion, error is CancellationError {
                 // Reset progress when cancelled
                 progress = Progress()
             }
             throw error
         }
     }
 }
+
+private actor WhisperKitTranscriptionWorker {
```
chen-argmax: please try break this down into smaller files, and add unit test for the functions you write

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -737,7 +737,29 @@ open class WhisperKit {
         seekOffsets: [Int]? = nil,
         callback: TranscriptionCallback? = nil
     ) async -> [Result<[TranscriptionResult], Swift.Error>] {
-        var result = [Result<[TranscriptionResult], Swift.Error>]()
+        return await transcribeWithOptions(
+            audioArrays: audioArrays,
+            decodeOptionsArray: decodeOptionsArray,
+            seekOffsets: seekOffsets,
+            callback: callback,
+            parentProgress: nil
```
chen-argmax, line 745: what does passing nil mean here? please add document in the private override funciton

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -748,82 +770,180 @@ open class WhisperKit {
                 return [.failure(WhisperError.transcriptionFailed("The number of audio arrays and seek offset indexes must be balanced."))]
             }
         }
+        
+        guard !audioArrays.isEmpty else {
+            return []
```
chen-argmax, line 783: add a log when this happens

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -1039,19 +1195,197 @@ open class WhisperKit {
...
+        
+        if let decodeOptions, decodeOptions.verbose {
+            transcribeTaskResult.logTimings()
+        }
+        
+        return [transcribeTaskResult]
+    }
+}
+
+private final class SharedTranscribeTaskDependencies: @unchecked Sendable {
```
chen-argmax: can this be an actor?

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -906,6 +1026,22 @@ open class WhisperKit {
...
+        return try await transcribe(
+            audioArray: audioArray,
+            decodeOptions: decodeOptions,
+            callback: callback,
+            segmentCallback: segmentCallback,
+            rootProgress: progress
+        )
+    }
+    
+    fileprivate func transcribe(
```
chen-argmax: why file private?

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -748,82 +770,180 @@ open class WhisperKit {
...
+        return await transcribeSequentially(
+            audioArrays: audioArrays,
+            decodeOptionsArray: decodeOptionsArray,
+            seekOffsets: seekOffsets,
+            callback: callback,
+            batchProgress: batchProgress
+        )
+    }
+    
+    private func transcribeSequentially(
```
chen-argmax: why do we need this? can we just add the addtiioan logic to `transcribe`?

Sources/WhisperKit/Core/WhisperKit.swift
naykutguven: Unfortunately it can't be an actor because we pass non-Sendable properties of non-Sendable `WhisperKit` here. We would be crossing an isolation boundary from WhisperKit's (nonisolated) isolation domain to actor's isolation domain and non-Sendable types aren't allowed to do that.

Sources/WhisperKit/Core/WhisperKit.swift
naykutguven: missed this one after code generation. removed now.

Sources/WhisperKit/Core/WhisperKit.swift
naykutguven: Done

Sources/WhisperKit/Core/WhisperKit.swift
naykutguven: moved to separate files

naykutguven: @chen-argmax added unit tests as well

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -737,7 +737,29 @@ open class WhisperKit {
...
+        decodeOptionsArray: [DecodingOptions?],
+        seekOffsets: [Int]?,
+        callback: TranscriptionCallback?,
+        parentProgress: Progress?
+    ) async -> [Result<[TranscriptionResult], Swift.Error>] {
+        if modelState != .loaded {
+            do {
+                try await loadModels()
+            } catch {
+                return [.failure(error)]
```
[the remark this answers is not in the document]
naykutguven: I think this is fine as is. wdyt @chen-argmax @ZachNagengast ?

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -1039,19 +1195,197 @@ open class WhisperKit {
...
+    ) {
+        self.currentTimings = currentTimings
+        self.audioProcessor = audioProcessor
+        self.audioEncoder = audioEncoder
+        self.featureExtractor = featureExtractor
+        self.segmentSeeker = segmentSeeker
+        self.textDecoder = textDecoder
+        self.tokenizer = tokenizer
+        self.voiceActivityDetector = voiceActivityDetector
+    }
```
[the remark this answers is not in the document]
naykutguven: This is the risk we talked about @chen-argmax . the concrete implementations are thread safe but if clients inject their own non-thread safe implementations, they may encounter data races. 

Making WhisperKit Sendable requires extensive locking or a queue **and/or** making its dependencies Sendable which is a huge constraint/responsibility on conforming classes.


 cc: @ZachNagengast

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -748,82 +770,180 @@ open class WhisperKit {
                 return [.failure(WhisperError.transcriptionFailed("The number of audio arrays and seek offset indexes must be balanced."))]
             }
         }
```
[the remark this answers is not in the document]
naykutguven: I think this is fine as is. wdyt @chen-argmax @ZachNagengast ?

Sources/WhisperKit/Core/WhisperKit.swift
naykutguven: Added documentation

Sources/WhisperKit/Core/WhisperKit.swift
naykutguven: we can but I think it keeps the intention cleaner and this method much less crowded. It is already around 100 lines. I prefer keeping functions up to 50 to 100 lines but I can move it inside if you prefer that?

Sources/WhisperKit/Core/TranscribeTaskDependencies.swift
```diff
@@ -0,0 +1,52 @@
+//
+//  File 2.swift
+//  whisperkit
+//
+//  Created by Aykut Güven on 18.02.26.
```
chen-argmax, line 5: please explan what this class does and why it's needed, ditto to other empty documented files

Sources/WhisperKit/Core/TranscriptionWorker.swift
```diff
@@ -0,0 +1,138 @@
+//  For licensing see accompanying LICENSE.md file.
+//  Copyright © 2024 Argmax, Inc. All rights reserved.
+
+import Foundation
+
+actor TranscriptionWorker {
```
chen-argmax, line 6: have you done any performance benchmark with the actor pool?

Sources/WhisperKit/Core/WhisperKit.swift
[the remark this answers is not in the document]
chen-argmax: have you tested the behavior before the refactor? were we returning multiple error results or one error result before? I think we should mirror the before behavior

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -737,7 +737,33 @@ open class WhisperKit {
...
+            decodeOptionsArray: decodeOptionsArray,
+            seekOffsets: seekOffsets,
+            callback: callback,
+            parentProgress: nil
+        )
+    }
+    
+    /// Internal batch transcription entry point.
+    ///
+    /// - Parameter parentProgress: If `nil`, creates a new batch-local `Progress` root (default public behavior).
```
chen-argmax, line 751: document all parameters.

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -737,7 +737,33 @@ open class WhisperKit {
...
+    ///   If non-`nil`, attaches each batch/chunk as children of that existing parent (used by recursive chunking paths).
+    private func transcribeWithOptions(
+        audioArrays: [[Float]],
+        decodeOptionsArray: [DecodingOptions?],
+        seekOffsets: [Int]?,
+        callback: TranscriptionCallback?,
+        parentProgress: Progress?
+    ) async -> [Result<[TranscriptionResult], Swift.Error>] {
+        if modelState != .loaded {
```
chen-argmax, line 760: why do we check this here? it wasn't checked before right?

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -748,82 +770,180 @@ open class WhisperKit {
...
+        let requestedWorkerCount: Int
+        if concurrentWorkerCount == 0 {
+            // Auto mode: bound isolated workers to CPU count to avoid loading one model pipeline per input.
+            let autoWorkerCount = max(1, ProcessInfo.processInfo.activeProcessorCount)
+            requestedWorkerCount = min(audioArrays.count, autoWorkerCount)
+        } else {
+            requestedWorkerCount = min(concurrentWorkerCount, audioArrays.count)
+        }
+        let batchProgress = parentProgress ?? Progress(totalUnitCount: Int64(audioArrays.count))
+        batchProgress.totalUnitCount = max(batchProgress.totalUnitCount, Int64(audioArrays.count))
```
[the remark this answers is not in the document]
chen-argmax: this seems a valid concern

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -748,82 +774,184 @@ open class WhisperKit {
...
+        defer {
+            batchProgress.completedUnitCount = batchProgress.totalUnitCount
```
chen-argmax: what does this line do?

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -748,82 +774,184 @@ open class WhisperKit {
...
+        }
+        
+        if requestedWorkerCount > 1 {
```
chen-argmax, line 799: this branching doesn't exit before right?

before we cut audioArray into number of concurrentWorkers, if it's one then audioArray is not cut. There was not special handling for no workers.

I think we would want to have a similar combined structure where 1 thread vs multiple threads share the same transcribe logic. do you don't need a dedicated `transcribeSequentially` and line 879

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -748,82 +774,184 @@ open class WhisperKit {
...
+    }
+    
+    private func transcribeSequentially(
+        audioArrays: [[Float]],
+        decodeOptionsArray: [DecodingOptions?],
+        seekOffsets: [Int]?,
+        callback: TranscriptionCallback?,
+        batchProgress: Progress
+    ) async -> [Result<[TranscriptionResult], Swift.Error>] {
+        let defaultSegmentDiscoveryCallback = self.segmentDiscoveryCallback
```
chen-argmax, line 859: this function seems copypasted from the branching logic above, can we somehow make them fit into a special case where only one actor is created?

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -748,82 +774,184 @@ open class WhisperKit {
...
+                result.append(.failure(error))
+            }
+        }
+        
         return result
     }
+    
+    private func createSharedTranscriptionWorkers(count: Int) -> [TranscriptionWorker] {
+        guard count > 0 else {
+            return []
```
chen-argmax, line 910: please log error for unexpected cases like this

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -748,82 +770,180 @@ open class WhisperKit {
...
+            featureExtractor: featureExtractor,
+            segmentSeeker: segmentSeeker,
+            textDecoder: textDecoder,
+            tokenizer: tokenizer,
+            voiceActivityDetector: voiceActivityDetector
+        )
+        return (0..<count).map { _ in
+            WhisperKitTranscriptionWorker(dependencies: dependencies)
+        }
+    }
```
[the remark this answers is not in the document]
chen-argmax: this is a valid concern imo

Sources/WhisperKit/Core/TranscribeTaskDependencies.swift
```diff
@@ -0,0 +1,52 @@
...
+        self.audioProcessor = audioProcessor
+        self.audioEncoder = audioEncoder
+        self.featureExtractor = featureExtractor
+        self.segmentSeeker = segmentSeeker
+        self.textDecoder = textDecoder
+        self.tokenizer = tokenizer
+        self.voiceActivityDetector = voiceActivityDetector
+    }
+
+    func makeTranscribeTask(progress: Progress) -> TranscribeTask {
```
chen-argmax, line 40: `WhisperKit.setupTranscribeTask` and `WhisperKit.runTranscribeTask` should be the only place where TranscribeTask is created/executed, because our pro instance extends WhisperKit's function sand alter the behavior

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -748,82 +774,184 @@ open class WhisperKit {
...
-                                callback: batchedAudioCallback,
-                                segmentCallback: batchedSegmentCallback
-                            )
-                            // Return the successful transcription result with its index
-                            return [(index: audioIndex, result: .success(transcribeResult))]
-                        } catch {
-                            // Return the failure result with its index in case of an error
-                            return [(index: audioIndex, result: .failure(error))]
-                        }
+                        await worker.transcribe(
```
chen-argmax: ditto to the comment in the work class itself, worker shouldn't create/run task within, they need to be overridble by sub classes of WhisperKit

Sources/WhisperKit/Core/WhisperKit.swift
```diff
@@ -906,6 +1034,22 @@ open class WhisperKit {
...
+        return try await transcribe(
+            audioArray: audioArray,
+            decodeOptions: decodeOptions,
+            callback: callback,
+            segmentCallback: segmentCallback,
+            rootProgress: progress
+        )
+    }
+    
+    private func transcribe(
```
chen-argmax, line 1022: this seems to be created for `transcribeSequentially`, which can be removed if we merge 1vsmultiple threads cases

Sources/WhisperKit/Core/WhisperKit.swift
chen-argmax, line 1118: ditto re. this function is overridden in the pro subclass.

changing its behavior would break our override

naykutguven:
based on the comments and the methods overridden by the pro SDK like setupTranscribeTask (thanks @chen-argmax  for the heads up and the comments), actor worker pool approach isn't the way to go. I am going to close the PR and work on making WhisperKit Sendable with locks or queues. This should also allow WhisperKit instances to be called on multiple threads which wasn't guaranteed in the worker pool approach.

PR #424, opened 2026-02-19
fix: replace force-unwraps with guards in TextDecoder.decodeText

### LINKED ISSUE #414: TextDecoder.decodeText force-unwraps cause crashes during repeated live transcription

yw5aj:
Fixes #414

Two force-unwraps in `decodeText` crash the app when decoder state degrades during sustained rapid transcription calls (e.g., live transcription preview at ~5 calls/sec):

- **Line 747**: `decoderInputs.initialPrompt.last!` — crashes when `initialPrompt` is empty
- **Line 829**: `decoderOutput.logits!` — crashes when logits is nil

This PR replaces them with guard statements that throw descriptive errors, allowing callers to handle the failure gracefully instead of crashing the entire app.

### Minimal diff — no behavioral change for normal operation
The guard statements only trigger when the values are unexpectedly nil. Normal transcription flow is unaffected.

### DIFF
--- a/Sources/WhisperKit/Core/TextDecoder.swift
+++ b/Sources/WhisperKit/Core/TextDecoder.swift
@@ -744,7 +744,10 @@ open class TextDecoder: TextDecoding, WhisperMLModel {
         let prefilledIndex = decoderInputs.cacheLength[0].intValue
         let initialPromptIndex = decoderInputs.initialPrompt.count
         var currentTokens: [Int] = decoderInputs.initialPrompt
-        var nextToken: Int = decoderInputs.initialPrompt.last!
+        guard let firstToken = decoderInputs.initialPrompt.last else {
+            throw WhisperError.decodingFailed("initialPrompt is empty")
+        }
+        var nextToken: Int = firstToken
         var logProbs: [Float] = Array(repeating: 0, count: currentTokens.count)
 
         // Logits filters
@@ -826,7 +829,10 @@ open class TextDecoder: TextDecoding, WhisperMLModel {
             let nonInferenceStartTime = Date()
 
             // Update predicted token as current
-            var logits = decoderOutput.logits!
+            guard let logits_ = decoderOutput.logits else {
+                throw WhisperError.decodingLogitsFailed("Decoder output logits are nil")
+            }
+            var logits = logits_
             for filter in logitsFilters {
                 logits = filter.filterLogits(logits, withTokens: currentTokens)
             }


ZachNagengast:
Interesting, third PR this week for this specific issue #420 #417. I see you are the author of the original issue, do you have any relation to the other two authors?

yw5aj:
No relation — we're an independent team that hit the same crash during live transcription. Glad to see others confirming the issue. Happy to defer to whichever PR you prefer to merge\!

ZachNagengast:
👍󠄳󠅜󠅟󠅣󠅕

achyutbenz19:
Took an independent look at the diff. A couple of things that might be useful.

**1. Rebase is clean.** The branch is on `2870b46`, and main has moved 6 commits ahead, most notably #456 (the package rename) and #455 (the swift-transformers to in-tree Hub/Tokenizers switch). `git rebase main` applies cleanly with no conflicts. Post-rebase, the patch is a pure +8/-2 on `Sources/WhisperKit/Core/TextDecoder.swift`. Confirmed with Swift 6.2 / Xcode 26 on Apple Silicon: the rebased HEAD builds, and the model-free TextDecoder tests (`testTextDecoderMLMultiArrayInputType`, `testTextDecoderMLMultiArrayOutputType`, `testTextDecoderMLMultiArrayOutputTypeWithNilValues`, `testDecodingCacheInitialization`, `testDecodingCacheWithNilValues`, `testMLMultiArrayConformsToTextDecoderTensorType`) pass 6/6 on both `main` and the rebased commit. Commands:

```
git fetch origin
git checkout fix/safe-unwrap-text-decoder
git rebase origin/main
```

**2. Adjacent pattern, noting for your judgment.** Lines 843 and 844 of current `main` (`sampleResult.tokens.last!`, `sampleResult.logProbs.last!`) are the same shape as the two unwraps this PR fixes. #420 covers them, this PR does not. Not suggesting you expand scope, just flagging in case it changes the merge calculus.

Disclosure: I am an AI assistant (Anthropic's Claude) helping a user contribute to WhisperKit. I did not reproduce the crash deterministically, it depends on rare runtime states that a file fixture will not reliably hit, so this is build and test signal only, not before / after empirical evidence.

PR #425, opened 2026-02-19
Add TTSKit with Qwen3-TTS support

ZachNagengast:
WhisperKit is expanding into text-to-speech!

TTSKit adds a new library for on-device text-to-speech using Core ML-accelerated [Qwen3-TTS](https://github.com/QwenLM/Qwen3-TTS) models (CustomVoice 0.6B and 1.7B in this first release) with real-time streaming playback on Apple Silicon. In this first PR, we're introducing the library into the WhisperKit package (WhisperKit will be renamed to reflect the new multi-Kit nature of Argmax Open-source SDK) as an optional import to add real-time TTS capabilities with a state-of-the-art open-source model, either on its own or as a complement to WhisperKit speech-to-text. 

This PR is still in the final phases of development, but here are a few highlights:

#### TTSKit Library
- Download, load, generate, and stream playback in ~3 lines of code.
- Protocol-based component architecture (6 swappable Core ML components: `TextProjecting`, `CodeEmbedding`, `MultiCodeEmbedding`, `CodeDecoding`, `MultiCodeDecoding`, `SpeechDecoding`) for plugging in new model backends.
- Qwen3-TTS implementation with 9 built-in voices, 10 languages, and style instruction support (1.7b variant only).
- Automatic text chunking for long-form generations with concurrent chunk generation and cross-fade stitching.
- Adaptive streaming playback (`TTSPlaybackStrategy.auto`) that measures first-step latency to pre-buffer just enough audio.
- Seedable RNG for reproducible generation.
- WAV and M4A (AAC) audio export

Example usage playing audio in real-time out of the default speaker: 
```swift
    let ttsKit = try await TTSKit()
    try await ttsKit.playSpeech(text: "Hello from TTSKit!")
```

#### New target: ArgmaxCore
- Extracted a shared target with various utilities from WhisperKit so TTSKit can share them without depending on it directly

#### CLI
- For now we plan to deploy this as a new command on whisperkit-cli `tts` that can be used like this:
    - `swift run whisperkit-cli tts --text "Hello from TTSKit" --play`
    - Full control over speaker, language, model, style instruction, temperature, chunking, compute units, and seed.

#### TTSKit Example app
- macOS and iOS example app with model management, real-time waveform visualization, generation history persisted as M4A files, and more. Use this as a quick way to try it out!

ttskit-example-app

#### Roadmap
We plan to continue to add support for state-of-the-art models and improve inference latency for TTSKit over the next few weeks. The immediate follow-up is the voice cloning feature from Qwen3-TTS and a 2x reduction in time-to-first-byte (TTFB) so this on-device project achieves a consistent sub-100 ms, providing a latency edge over cloud deployments of the same model. In the meantime, we encourage anyone reading this to check out this PR, give it a spin, and let us know how it goes!


Examples/TTS/TTSKitExample/TTSKitExample/ViewModel.swift
```diff
@@ -0,0 +1,1076 @@
...
+    @AppStorage("codeDecoderComputeUnits") var codeDecoderComputeUnitsRaw: Int = MLComputeUnits.cpuAndNeuralEngine.rawValue
+    @AppStorage("multiCodeDecoderComputeUnits") var multiCodeDecoderComputeUnitsRaw: Int = MLComputeUnits.cpuAndNeuralEngine.rawValue
+    @AppStorage("speechDecoderComputeUnits") var speechDecoderComputeUnitsRaw: Int = MLComputeUnits.cpuAndNeuralEngine.rawValue
+}
+
+// MARK: - View Model
+
+@MainActor
+@Observable
+final class ViewModel: @unchecked Sendable {
```
chen-argmax, line 147: I would break this down to smaller viewmodels if it goes too long, e.g DownloadViewModel vs. TTSViewModel

Examples/WhisperAX/WhisperAX.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,5 +1,5 @@
 {
```
chen-argmax, line 1: why do we need to change this?

Sources/ArgmaxCore/ConcurrencyUtilities.swift
```diff
@@ -0,0 +1,86 @@
...
+
+import Foundation
+import os.lock
+
+// MARK: - Unfair Lock
+
+/// Thin wrapper around `os_unfair_lock` that exposes a Swift-friendly
+/// `withLock` helper. This lock is non-reentrant and optimized for low
+/// contention, matching the semantics of Core Foundation's unfair lock.
+public final class UnfairLock: @unchecked Sendable {
```
chen-argmax, line 23: I think we want to make this class name generic for future proof with swift6, seems `os_unfair_lock` is not the recommended way to lock in swift 6.
probably rename it `Mutext` so we can reimp it with actual Swift.Mutext later
now
```
public final class Mutex: @unchecked Sendable {
    private let lock = OSAllocatedUnfairLock()

    public init() {}

    @inlinable
    public func withLock<T>(_ body: () throws -> T) rethrows -> T {
        try lock.withLock(body)
    }
}


later
```
public final class Mutex<Value>: Sendable {
    private let mutex: Swift.Mutex<Value>

    public init(_ value: Value) {
        self.mutex = Mutex(value)
    }

    public func withLock<T>(_ body: (inout Value) throws -> T) rethrows -> T {
        try mutex.withLock(body)
    }
}
```

File: Sources/ArgmaxCore/ConcurrencyUtilities.swift [hunk 4]
```diff
@@ -0,0 +1,86 @@
+//  For licensing see accompanying LICENSE.md file.
+//  Copyright © 2024 Argmax, Inc. All rights reserved.
```
chen-argmax, line 2: 2026, ditto to others

Sources/ArgmaxCore/MLModelExtensions.swift
chen-argmax, line 1: should we consider adding another package under ArgmaxCore? like `ArgmaxCore/CoreML`

Sources/TTSKit/Qwen3TTS/Qwen3CodeDecoder.swift
```diff
@@ -0,0 +1,155 @@
...
+
+// MARK: - Implementation
+
+/// Autoregressive codec-0 decoder backed by a CoreML model.
+///
+/// Thread safety: mutable state (`model`, dimension properties) is set once during
+/// `loadModel()` and read-only thereafter. `MLModel.prediction()` is thread-safe.
+/// Per-generation `MLState` is created via `makeState()` and owned by each task,
+/// never stored on this shared instance.
+public class TTSCodeDecoder: CodeDecoding, @unchecked Sendable {
```
chen-argmax: do we want to make this open for pro overriding? ditto to other components

Sources/TTSKit/Utilities/AudioOutput.swift
```diff
@@ -0,0 +1,519 @@
...
+/// 2. `setBufferDuration(_:)` - configures threshold (call after start).
+/// 3. `enqueueAudioChunk(_:)` - pushes frames through the buffer/tail pipeline.
+/// 4. `stopPlayback()` - commits the tail with fade-out, waits, tears down.
+///
+/// Thread safety: used only from `TTSKit.playSpeech()` which forces sequential mode
+/// (concurrency=1). `TTSKit` ensures serialized access.
+/// Subclassing is intentionally not supported. Use the `TTSKit` component override
+/// mechanism (`TTSKitConfig`) to swap in an alternative audio backend.
+public class TTSAudioOutput: @unchecked Sendable {
+    private var audioEngine: AVAudioEngine?
```
chen-argmax, line 39: are these not qwen specific? imo the neutral TTS apis should only have model agnostic interfaces

Sources/TTSKit/TTSAudioOutput.swift
```diff
@@ -0,0 +1,519 @@
...
+
+    /// Save PCM samples as an M4A (AAC) file with optional embedded metadata.
+    ///
+    /// Pass `[AVMetadataItem]` built by the caller - TTSKit itself has no opinion
+    /// about what metadata to embed, keeping this layer generic.
+    ///
+    /// - Note: `async` because `AVAssetExportSession.export()` is asynchronous.
+    @available(watchOS, unavailable, message: "AVAssetExportSession is not available on watchOS")
+    @MainActor
+    public static func saveAudioAsM4A(
```
chen-argmax: This seems to belong to ArgmaxCore.AudioUtils, dito to other static audio utils

Sources/TTSKit/TTSGenerating.swift
```diff
@@ -0,0 +1,37 @@
...
+///
+/// `TTSKit` creates tasks via `makeTask(progress:)` and calls `run(...)` on each one.
+/// Each task instance is independent - fresh KV caches, own RNG - so multiple tasks
+/// can run concurrently without data races.
+///
+/// The interface uses plain `String` for `voice` and `language` to stay model-agnostic.
+/// Each implementation maps these to its internal representation (e.g. token IDs).
+///
+/// **Reference implementation:** `TTSGenerateTask` in `Sources/TTSKit/Qwen3TTS/Qwen3TTSGenerateTask.swift`.
+public protocol TTSGenerating: Sendable {
```
chen-argmax: ditto to the above qwen specific question, I think this and TTSKit.swift should be the only file that's not under TTSKit/Qwen3TTS

Sources/TTSKit/TTSKit.swift
```diff
@@ -0,0 +1,600 @@
...
+        }
+    }
+
+    // MARK: - Download
+
+    /// Download Qwen3 TTS models from HuggingFace Hub.
+    ///
+    /// Downloads only the files matching the configured component variants.
+    /// Files are cached locally by the Hub library.
+    open class func download(
```
chen-argmax, line 332: should we decouple model download from TTSKit? ArgmaxCore could provide a downloader for this

Sources/TTSKit/TTSModels.swift
```diff
@@ -0,0 +1,560 @@
+//  For licensing see accompanying LICENSE.md file.
+//  Copyright © 2026 Argmax, Inc. All rights reserved.
+
+import Accelerate
+@_exported import ArgmaxCore
```
chen-argmax: why @_exported?

Tests/TTSKitTests/TTSKitIntegrationTests.swift
```diff
@@ -0,0 +1,278 @@
...
+        let tts = try await makeCachedTTS(seed: 42)
+        let result = try await tts.generateSpeech(
+            text: "Hello, this is a basic smoke test of the WhisperKit TTS pipeline.",
+            speaker: .ryan,
+            language: .english,
+            options: TTSGenerationOptions(temperature: 0.9, topK: 50, maxNewTokens: 245)
+        )
+
+        XCTAssertGreaterThan(result.audio.count, 0, "Audio samples should be non-empty")
+        XCTAssertGreaterThan(result.audioDuration, 1.0, "Expect at least 1s of speech")
```
chen-argmax, line 44: will seed guarantee the audio length is always deterministic?

Tests/TTSKitTests/TTSKitUnitTests.swift
```diff
@@ -0,0 +1,668 @@
+//  For licensing see accompanying LICENSE.md file.
+//  Copyright © 2024 Argmax, Inc. All rights reserved.
+
+import ArgmaxCore
```
chen-argmax, line 4: I think we would want to break these test down to isolated class test.

e.g1 `TTSKitTest.swift` that injects a Config with mocked components, and verify 
`TTSKitTest.generateSpeech` interacts with the components correctly, tasks created etc.

e.g2 `Qwen3TTSGenerateTaskTest.swfit` that inejcts mocked components. verify `run` interacts with them correctly

Sources/TTSKit/Qwen3TTS/Qwen3TTSGenerateTask.swift
```diff
@@ -0,0 +1,408 @@
...
+/// runs a complete prefill + autoregressive decode cycle.
+///
+/// Conforms to `TTSGenerating` so it can be returned by `TTSKit.makeTask(progress:)`
+/// and consumed by `TTSKit`'s generic orchestration layer.
+///
+/// Thread safety: all stored properties are `let` (immutable after init). Each task
+/// owns its own sampler (derived seed) so concurrent tasks don't share RNG state.
+/// Model components are shared read-only references - `MLModel.prediction()` is
+/// thread-safe. The class is `@unchecked Sendable` to permit `open` subclassing.
+open class TTSGenerateTask: @unchecked Sendable, TTSGenerating {
```
chen-argmax: Should the class be renamed to `Qwen3TTSGenerateTask` ? ditto to other files under Qwen3TTS

chen-argmax (changes requested):

Tests/TTSKitTests/TTSKitIntegrationTests.swift
ZachNagengast: Yup, apple docs recommend using this method https://developer.apple.com/documentation/swift/randomnumbergenerator#Conforming-to-the-RandomNumberGenerator-Protocol

Sources/TTSKit/TTSKit.swift
ZachNagengast: Yep have some todos relating to this

Sources/ArgmaxCore/ConcurrencyUtilities.swift
ZachNagengast: This was brought over from https://github.com/argmaxinc/WhisperKit/blob/main/Sources/WhisperKit/Utilities/Concurrency.swift

Sources/TTSKit/TTSAudioOutput.swift
ZachNagengast: Have a todo for this as well 👍

Sources/TTSKit/Utilities/AudioOutput.swift
ZachNagengast: This is intended to be generic yes.

Examples/WhisperAX/WhisperAX.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
ZachNagengast: Had an old swift-transformers resolved

Sources/ArgmaxCore/ConcurrencyUtilities.swift
```diff
@@ -0,0 +1,86 @@
...
+        return try body()
+    }
+}
+
+// MARK: - Property Lock
+
+/// Serializes access to a value with an `os_unfair_lock` so mutation stays
+/// thread-safe. Useful for properties on types marked `@unchecked Sendable`.
+@propertyWrapper
+public struct PropertyLock<Value: Codable & Sendable>: Sendable, Codable {
```
naykutguven, line 55: TLDR; @ZachNagengast @chen-argmax guys, this doesn't make reference or value type properties truly thread safe.

I was playing around with this and trying to move `Sendable` and `Codable` conformances outside. Did some verifications on the current implementation and mine. Ran the snippet below with different variations
1. Reference type property `Ref`
2. Value type property `Ref`
3. Plain property of type `Int`

None of them was safe. Locking accessors isn't enough. We need to wrap mutations with locks

```swift
final class Ref: Codable, @unchecked Sendable {
    var count: Int

    init(count: Int = 0) { self.count = count }

    enum CodingKeys: String, CodingKey { case count }

    required init(from decoder: Decoder) throws {
        let c = try decoder.container(keyedBy: CodingKeys.self)
        self.count = try c.decode(Int.self, forKey: .count)
    }

    func encode(to encoder: Encoder) throws {
        var c = encoder.container(keyedBy: CodingKeys.self)
        try c.encode(count, forKey: .count)
    }
}

final class Holder: @unchecked Sendable {
    @TranscriptionPropertyLock var ref = Ref()
}

@main
struct Main {
    static func main() async {
        let workers = max(2, ProcessInfo.processInfo.activeProcessorCount * 2)
        let perWorker = 50_000
        let expected = workers * perWorker

        print("workers=\(workers), perWorker=\(perWorker), expected=\(expected)")

        for run in 1...10 {
            let holder = Holder()

            await withTaskGroup(of: Void.self) { group in
                for _ in 0..<workers {
                    group.addTask {
                        for _ in 0..<perWorker {
                            holder.ref.count += 1
                        }
                    }
                }
            }

            let final = holder.ref.count
            print("run \(run): expected=\(expected) actual=\(final)")
        }
    }
}
```

naykutguven: The approach used in [AudioProcessor PR](https://github.com/argmaxinc/WhisperKit/pull/412/changes#diff-4d25eabe13b7cc336500ee7b86380609a150a4f181fd2a626591279e49fbce2eR208) (also the WhisperKit PR) works. Snippet below:

```swift
import Foundation
import os.lock

@usableFromInline
final class UnfairLock: @unchecked Sendable {
    @usableFromInline
    var lock = os_unfair_lock()

    @inlinable
    func withLock<T>(_ body: () throws -> T) rethrows -> T {
        os_unfair_lock_lock(&lock)
        defer { os_unfair_lock_unlock(&lock) }
        return try body()
    }
}

final class Ref {
    var count = 0
}

final class HolderInt: @unchecked Sendable {
    private let stateLock = UnfairLock()
    private var countStorage = 0

    var count: Int {
        get {
            stateLock.withLock { countStorage }
        }
        set {
            stateLock.withLock { countStorage = newValue }
        }
    }

    func increment() {
        stateLock.withLock {
            countStorage += 1
        }
    }
}

final class HolderRef: @unchecked Sendable {
    private let stateLock = UnfairLock()
    private let refStorage = Ref()

    var refCount: Int {
        stateLock.withLock { refStorage.count }
    }

    func incrementRef() {
        stateLock.withLock {
            refStorage.count += 1
        }
    }
}

@main
struct Main {
    static func main() async {
        let workers = max(2, ProcessInfo.processInfo.activeProcessorCount * 2)
        let perWorker = 50_000
        let expected = workers * perWorker

        print("[Int] workers=\(workers), perWorker=\(perWorker), expected=\(expected)")
        for run in 1...10 {
            let holder = HolderInt()
            await withTaskGroup(of: Void.self) { group in
                for _ in 0..<workers {
                    group.addTask {
                        for _ in 0..<perWorker {
                            holder.increment()
                        }
                    }
                }
            }
            let final = holder.count
            print("[Int] run \(run): expected=\(expected) actual=\(final)")
        }

        print("[Ref] workers=\(workers), perWorker=\(perWorker), expected=\(expected)")
        for run in 1...10 {
            let holder = HolderRef()
            await withTaskGroup(of: Void.self) { group in
                for _ in 0..<workers {
                    group.addTask {
                        for _ in 0..<perWorker {
                            holder.incrementRef()
                        }
                    }
                }
            }
            let final = holder.refCount
            print("[Ref] run \(run): expected=\(expected) actual=\(final)")
        }
    }
}
```

zaidbren:
I am trying to run the 1.7B model on macbook air m1, and although the 0.6B version worked fine, in the 1.7B, It first specialize the model for the device, than loading and when it was generating, it stopped and throws this error :- Unable to compute the prediction using ML Program. It can be an invalid input data or broken/unsupported model.

Sources/ArgmaxCore/ConcurrencyUtilities.swift
chen-argmax: This is a valid concern, essentially if the property wrapped property has another property, read/write wont' be thread safe.

e.g this is thread safe
```
holder.ref = otherRef
```

this is not thread safe
```
holder.ref.count += 1
```

@ZachNagengast we may want to add document for this wrapper.

naykutguven: I think it isn't safe to use it for pure value type properties  e.g. `Int` either. we probably need to use `_modify` instead of `set`.

I am checking these resources:
- https://www.linkedin.com/pulse/migrating-swift-6-dont-use-unchecked-sendable-until-you-kolyadin-mflee/
- https://github.com/artemkolyadin/swift-threadsafe-macros
- https://www.youtube.com/watch?v=035WscXr7Xo&t=937s (Russian so I used auto translated CC) cc: @ZachNagengast

ZachNagengast: Making a note in this PR but will leave the fix to a followup 👍

PR #428, opened 2026-02-22
Fix prompt tokens causing empty transcription output

sborisov88:
## Summary

When `promptTokens` are provided in `DecodingOptions` (e.g., via the `prompt` parameter in the OpenAI-compatible server API), transcription returns empty text.

**Root cause:** When `promptTokens` are set, the prefill KV cache is disabled ([known TODO on line 354](https://github.com/argmaxinc/WhisperKit/blob/664e1b5/Sources/WhisperKit/Core/TextDecoder.swift#L354)). This causes the decoding loop to start at `tokenIndex=0` with `startOfPreviousToken` as input. The model then either:
1. Predicts EOT → `sampleResult.completed = true` → loop breaks immediately
2. Produces a low-confidence prediction → `firstTokenLogProbThreshold` (-1.5 default) triggers → loop breaks

Both paths result in empty transcription output.

## Reproduction

```bash
# Start WhisperKit server with any model
whisperkit-cli serve --model-path <path> --port 8000

# Without prompt — works fine
curl -X POST http://localhost:8000/v1/audio/transcriptions \
  -F "file=@test.wav" -F "model=<model>" -F "language=ru"
# → {"text": "Привет, это тестовая запись..."}

# With prompt — returns empty text
curl -X POST http://localhost:8000/v1/audio/transcriptions \
  -F "file=@test.wav" -F "model=<model>" -F "language=ru" \
  -F "prompt=Kubernetes, Docker, RabbitMQ"
# → {"text": ""}
```

## Fix

Two changes in `TextDecoder.swift`:

1. **`isFirstToken`** now correctly points to the first actually decoded token after the prompt (`max(prefilledIndex, initialPromptIndex)`) instead of `tokenIndex == prefilledIndex` which fires at the first prompt token during prefill.

2. **`sampleResult.completed`** (EOT check) is skipped during the prefill phase. Since the model is being force-fed prompt tokens, its predictions during prefill are not meaningful for early stopping decisions.

## Test plan

- [x] Transcription without prompt still works correctly
- [x] Transcription with prompt now returns correct text (previously returned empty)
- [x] Tested with `whisper-large-v3-turbo` model via local server API

### DIFF
--- a/Sources/WhisperKit/Core/TextDecoder.swift
+++ b/Sources/WhisperKit/Core/TextDecoder.swift
@@ -764,7 +764,8 @@ open class TextDecoder: TextDecoding, WhisperMLModel {
 
             let isPrefill = tokenIndex < initialPromptIndex - 1 // Prefill stops at the last token of the initial prompt
             let isLastPrefillToken = tokenIndex == initialPromptIndex - 1
-            let isFirstToken = tokenIndex == prefilledIndex
+            let isInPrefillPhase = isPrefill || isLastPrefillToken // tokenIndex < initialPromptIndex
+            let isFirstToken = tokenIndex == max(prefilledIndex, initialPromptIndex) // First actually decoded token (after prompt)
 
             // Check if current index is part of the initial prompt
             if tokenIndex < initialPromptIndex {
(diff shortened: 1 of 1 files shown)


atiorh:
@sborisov88 Could you please add a non-trivial test case that gets fixed by this PR? (e.g. a short audio with a keyterm that the model gets wrong even with prompting but gets right after this fix)

a2they:
Thank you for this PR. You identified both failure paths here, including the `firstTokenLogProbThreshold` one, and that made everything that followed easier. I opened PR #514 building on your analysis with two small boundary tweaks: the EOT gate uses `!isPrefill` so that a real EOT at the first decoded token (for example silence) still ends the segment right away, and `isFirstToken` anchors at `initialPromptIndex - 1`, since the first decoded token is sampled one step earlier than `initialPromptIndex`. You are credited as a co-author. Thanks again!

PR #443, opened 2026-03-15
Fix data race in AudioProcessor: thread-safe audioSamples and audioEnergy

### LINKED ISSUE #442: Data race in AudioProcessor: audioEnergy/audioSamples accessed from multiple threads without synchronization
## Summary

`AudioProcessor` has data races on `audioEnergy` and `audioSamples` — they are written from the audio tap callback thread and read from arbitrary threads (e.g., main thread) without any synchronization.

## Details

In `AudioProcessor.swift`, `processBuffer()` is called from the audio engine tap's callback thread:

```swift
// ~line 907-917, called from audio tap callback thread
func processBuffer(_ buffer: [Float]) {
    audioSamples.append(contentsOf: buffer)          // write
    // ...
    self.audioEnergy.append(newEnergy)                // write
}
```

Meanwhile, `relativeEnergy` reads these from any thread:

```swift
// ~line 210, read from any thread
public var relativeEnergy: [Float] {
    return self.audioEnergy.map { $0.rel }
}
```

And `audioSamples` is also read externally:

```swift
public var audioSamples: ContiguousArray<Float> = []  // no synchronization
```

This is a Swift Concurrency data race — concurrent read/write on non-isolated, non-sendable mutable state.

## Impact

- When polling `relativeEnergy` from the main thread (e.g., for VAD-based end-of-speech detection), values can be stale, skipped, or inconsistent.
- In our case, this caused `AudioProcessor.isVoiceDetected()` to not work reliably when called from a `@MainActor`-isolated context while recording.
- Could potentially crash in optimized builds due to unsynchronized array mutations.

## Reproduction

1. Start recording with `audioProcessor.startRecordingLive()`
2. Poll `audioProcessor.relativeEnergy` from the main thread in a timer
3. Observe that values are inconsistent or empty

## Suggested Fix

Either:
- Make `AudioProcessor` an `actor`
- Or protect `audioEnergy` / `audioSamples` with a lock (e.g., `NSLock` or `os_unfair_lock`)

## Environment

- WhisperKit 0.17.0
- Xcode 26.3
- iOS 26.3 / macOS 15

shawnzhu:
## Summary

- Add `NSLock` to protect `audioSamples` and `audioEnergy` from concurrent access
- These properties are written from the audio tap callback thread (in `processBuffer`) and read from arbitrary threads (e.g., main thread polling `relativeEnergy` for VAD)
- Lock is held only for the minimum critical section — energy calculation and callback invocation happen outside the lock

## Problem

`audioSamples` and `audioEnergy` are plain `var` properties with no synchronization, causing data races when:
1. The audio tap callback writes via `processBuffer()`
2. Other threads read via `relativeEnergy`, `audioSamples`, or `purgeAudioSamples()`

This manifests as stale/inconsistent values when polling `relativeEnergy` from the main thread for voice activity detection, and can potentially crash in optimized builds.

## Test plan

- [x] All 194 existing unit tests pass
- [x] `swift build` succeeds
- [x] Manual test: poll `relativeEnergy` from main thread during recording — values should update consistently

Fixes #442

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/WhisperKit/Core/Audio/AudioProcessor.swift
+++ b/Sources/WhisperKit/Core/Audio/AudioProcessor.swift
@@ -204,11 +204,27 @@ public extension AudioProcessing {
 open class AudioProcessor: NSObject, AudioProcessing {
     private var lastInputDevice: DeviceID?
     public var audioEngine: AVAudioEngine?
-    public var audioSamples: ContiguousArray<Float> = []
-    public var audioEnergy: [(rel: Float, avg: Float, max: Float, min: Float)] = []
+
+    /// Lock protecting `_audioSamples` and `_audioEnergy` from concurrent access.
+    /// These properties are written from the audio tap callback thread and read
+    /// from arbitrary threads (e.g. main thread for VAD polling).
+    private let audioLock = NSLock()
+    private var _audioSamples: ContiguousArray<Float> = []
+    private var _audioEnergy: [(rel: Float, avg: Float, max: Float, min: Float)] = []
+
+    public var audioSamples: ContiguousArray<Float> {
+        get { audioLock.withLock { _audioSamples } }
+        set { audioLock.withLock { _audioSamples = newValue } }
+    }
+
+    public var audioEnergy: [(rel: Float, avg: Float, max: Float, min: Float)] {
+        get { audioLock.withLock { _audioEnergy } }
+        set { audioLock.withLock { _audioEnergy = newValue } }
+    }
+
     public var relativeEnergyWindow: Int = 20
     public var relativeEnergy: [Float] {
-        return self.audioEnergy.map { $0.rel }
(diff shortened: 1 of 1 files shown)


shawnzhu:
human here: manually tested and it works for me.

The use case is using Whisper's audio energy to detect speech pause.

shawnzhu:
@ZachNagengast would you be able to review this? You have the most context on the AudioProcessor code. Happy to address any feedback.

ZachNagengast:
@shawnzhu Thanks for the contribution, my main thought currently is that this is handled in https://github.com/argmaxinc/WhisperKit/pull/412, in somewhat of a different way. Would you be able to test out that PR with your implementation and see if it covers what you were aiming for with this? If not, we'll have to reconcile them.

PR #463, opened 2026-04-19
Expose speaker centroid embeddings on DiarizationResult

### LINKED ISSUE #457: Expose per-speaker embeddings in DiarizationResult
Hey,

to be honest i used Claude for this but it feels legit and would be a highly wanted improvement:

**Request**
The internal SpeakerEmbedding struct and the embedding vectors computed by SpeakerEmbedderModel are not surfaced through the public API. DiarizationResult currently only exposes timing segments ([SpeakerSegment]), but not the [Float] embedding vectors that Pyannote computes for each speaker cluster.

**Use case**
Cross-meeting speaker identification: after a user names a speaker in one meeting, their voice embedding can be stored in a profile and compared (cosine similarity) against speaker embeddings from future meetings to automatically pre-fill names — without any additional ML models.

The probe audio clips and profile linking infrastructure are already in place on the app side. The only missing piece is access to the embedding vectors.

**Request**
Please add the per-speaker (per-cluster) mean/centroid embeddings to DiarizationResult, for example:

```
public struct DiarizationResult: Sendable {
    // existing fields ...
    public var speakerEmbeddings: [Int: [Float]]  // speakerId → embedding vector
}
```

Or alternatively, expose a method on SpeakerKit / PyannoteDiarizer to compute an embedding for a given audio clip:

`public func embed(audioArray: [Float]) async throws -> [Float]`

Either form would be sufficient.

**Context**
- SpeakerEmbedding (internal struct) already holds the embedding: [Float] and pldaEmbedding: [Float]? per window
- SpeakerEmbedderModel.embed() already aggregates them per cluster during the diarization pipeline
- Surfacing the cluster-level centroid would add no computational overhead

leecrossley:
## Summary

Adds `speakerCentroidEmbeddings: [Int: [Float]]` to `DiarizationResult` so downstream consumers can match speakers across diarization runs without re-running the embedder.

Centroids are computed inside the clusterer, not in `postProcess`. `VBxClustering.cluster(...)` returns post-reassignment centroids along all three paths (VBx weighted, kMeans correction, AHC fallback) via one `centroidsFromAssignments(assignments: clusters, embeddings: all, clusterCount: kFinal)` pass after `clusterReassignment(...)`, so `speakerCentroidEmbeddings[k]` is the mean of the final cluster members of speaker `k`.

Closes #457.

## Motivation

We ship a privacy-first recording app that runs SpeakerKit diarization over many short audio chunks and needs to link speakers to the same person across chunks. Today, the per-window embeddings used internally for clustering are thrown away before `diarize(...)` returns, leaving no way to correlate the cluster ids in one result with those in another without running the embedder a second time over the whole chunk.

Exposing the cluster centroids is the smallest change that makes this possible: callers get one embedding per cluster, computed from the same data the clusterer already used. https://over.show

## Changes

- `Sources/SpeakerKit/Pyannote/SpeakerClustering.swift`: add `speakerCentroids: [Int: [Float]]` on `ClusteringResult` with default `[:]` so non-VBx conformers stay compatible.
- `Sources/SpeakerKit/Pyannote/VBxClustering.swift`: `cluster(...)` now returns `(clusters, linkageMatrix, centroids)`; after `clusterReassignment(...)` it runs one extra `centroidsFromAssignments(...)` pass so the surfaced map is the mean of the final cluster members across all three paths.
- `Sources/SpeakerKit/Pyannote/PyannoteDiarizer.swift`: `postProcess` accepts `speakerCentroids` and threads them into `DiarizationResult`; inline mean-pool loop removed.
- `Sources/SpeakerKit/DiarizationResult.swift`: `speakerCentroidEmbeddings` is `public private(set) var` with doc comments on raw embedder space, post-reassignment mean, threshold-free distance semantics, and Pyannote-only applicability. Adds `public func centroidCosineDistance(between:_:)` and `public func nearestSpeakerCentroid(to:)` for caller-side comparison.
- `Sources/SpeakerKit/Pyannote/SpeakerEmbedderModel.swift`: revert `SpeakerEmbedding` and its `embedding` field back to `internal`; nothing else in the PR needs them public now that the compute lives in the clusterer.
- `Tests/SpeakerKitTests/SpeakerCentroidEmbeddingsTests.swift`: unit tests across both centroid producers (`calculateCentroids` + `centroidsFromAssignments`) and integration tests on `VADAudio`/`VBxClustering`, including `testCentroidValuesMatchFinalAssignmentMean` which pins the surfaced value equals the mean of the final cluster members.

## Cost

The runtime addition is one final `O(N x D)` mean-pool on final assignments inside `VBxClustering.cluster(...)` (N embeddings, D=192). No performance harness is included in this PR.

## Test plan

- `swift test --filter SpeakerCentroidEmbeddingsTests`: 14 tests, 0 failures, 1 skip (`testCentroidCosineDistance_sameDiarization` skips on bundled single-speaker fixtures; the helper is separately covered by unit tests).
- `swift test --filter SpeakerKitTests`: 109 tests, 0 failures, 1 skip.
- `git diff --check`: clean.


ZachNagengast:
Thanks for the PR, could you please include tests for this in your PR? Particularly interested in any latency overhead this adds as well. Also the link to your downstream consumer may be a private repo, resolves to 404.

Sources/SpeakerKit/Pyannote/PyannoteDiarizer.swift
```diff
@@ -280,6 +280,27 @@ actor PyannoteDiarizerActor {
             return DiarizationResult(binaryMatrix: [], diarizationFrameRate: diarizationFrameRate)
         }
 
+        var centroidSums: [Int: [Float]] = [:]
+        var centroidCounts: [Int: Int] = [:]
+        for emb in speakerEmbeddings {
```
a2they: This is inconsistent how pyannote computes centroids. You should add centroids map or list to `ClusteringResult` to set during clustering and pass it into the post process function.

Sources/SpeakerKit/Pyannote/SpeakerEmbedderModel.swift
```diff
@@ -4,8 +4,8 @@
 import CoreML
 import ArgmaxCore
 
-struct SpeakerEmbedding {
-    let embedding: [Float]
+public struct SpeakerEmbedding {
```
a2they: why is this public?

Sources/SpeakerKit/DiarizationResult.swift
```diff
@@ -30,15 +30,17 @@ public struct DiarizationResult: Sendable {
     public let frameRate: Float
     public private(set) var segments: [SpeakerSegment]
     public var timings: (any DiarizationTimings)?
+    public var speakerCentroidEmbeddings: [Int: [Float]]
```
a2they: you should add a comment about embedding space (raw embedder vs plda vs normalied etc) and suggest how to compare distance or even better add utility function to compute distance between two centroids

a2they: Also, instead of `public var` I suggest `private(set) var` as it should only be set internally

a2they: Please also add a comment that this field is applicable only to pyannote backend (not all backends will have a centroid to set for each speaker)

a2they (changes requested):
Please add a unit test to make sure the correct cendroid is properly set

leecrossley:
thanks both, addressed + pushed.

**re 404**: overshow is a private commerical repo. usage:

```swift
let diarization = try await speakerKit.diarize(audioArray: audio)
let centroids = diarization.speakerCentroidEmbeddings
return SpeakerMatch(id: speakerId, embedding: centroids[speakerId])
```

we persist the centroid with each transcribed segment and cosine-match aroujnd chunks to keep speaker ids stable wthout re-running the embedder. https://over.show

**re latency**: centroid map now surfaces from `ClusteringResult` instead of being recomputed in `postProcess`. `VBxClustering.cluster(...)` returns post-reassignment centroids along all three paths (VBx weighted, kMeans correction, AHC fallback) via one `centroidsFromAssignments(assignments: clusters, embeddings: all, k: kFinal)` pass after `clusterReassignment(...)`. That's one extra O(N x D) mean-pool on the final assignments (N embeddings, D=192) - small next to the model pipeline, and the measured delta below confirms it's inside run-to-run noise.

VADAudio, `XCTClockMetric`, 20 iters, preload + warmup outside `measure`, same machine:

| branch | mean (ms) | RSD |
|---|---|---|
| main  | 305.7 | 0.995% |
| pr    | 306.3 | 0.772% |
| delta | +0.6 ms (+0.2%) | |

inside run-to-run noise (delta smaller than either rsd).

**re tests**: new `SpeakerCentroidEmbeddingsTests.swift` - 6 unit across both centroid producers, 5 integration (VADAudio + `VBxClustering`) incl. a post-reassignment centroid-value regression that pins the surfaced value equals the mean of the final cluster members. full `make test` green on the branch after `make download-speakerkit-models`.

happy to rebase/squash.

Sources/SpeakerKit/Pyannote/PyannoteDiarizer.swift
leecrossley: ah, catch, moved. `VBxClustering.cluster(...)` now returns postreassignment centroids along all three paths - computed nce via `centroidsFromAssignments(assignments: clusters, embeddings: all, k: kFinal)` after `clusterReassignment(...) ` so the surfaced value is the mean of the final members, not the pre-reassignment approximation. `postProcess` just threads `ClusteringResult.speakerCentroids` into `DiarizationResult`, no extra loop

Sources/SpeakerKit/Pyannote/SpeakerEmbedderModel.swift
leecrossley: reverted. nothing else in the PR needs it now the compute lives in the clusterer

leecrossley:
> Please add a unit test to make sure the correct cendroid is properly set

Added. unit coverage on both producers: calculateCentroids (main VBx path) and centroidsFromAssignments (kMeans correction + AHC fallback). integration testCentroidKeysSurviveClusterReassignment pins that every speaker id visible in final segments has a centroid after reassignment and testCentroidValuesMatchFinalAssignmentMean pins the centroid VALUE against the mean of final members - so "correct centroid is properly set" holds post-reassignment, not just pre-

Sources/SpeakerKit/DiarizationResult.swift
leecrossley: both done. doc comment: raw embedder output, unnormalised, pre-PLDA, post-reassignment mean. added:

```swift
public func centroidCosineDistance(between a: Int, _ b: Int) -> Float?
```

delegates to the existing internal `MathOps.cosineDistance(_:_:)` (vDSP, clamped to `[0, 2]`) so numerics match `MathOps.cosineDistanceMatrix` used by `clusterReassignment`. no new `Accelerate` import in `DiarizationResult.swift`. if you'd rather expose the helper itself, happy to mark `MathOps.cosineDistance` `public` in the same PR.

leecrossley: done, `public private(set) var` to match `segments`.

leecrossley: added. doc notes it's populated by `PyannoteDiarizer`; other backends may leave it `[:]`.

Tests/SpeakerKitTests/SpeakerCentroidEmbeddingsTests.swift
```diff
@@ -0,0 +1,379 @@
...
+        file: StaticString = #filePath,
+        line: UInt = #line
+    ) {
+        XCTAssertEqual(actual.count, expected.count, "Vector lengths differ", file: file, line: line)
+        for (i, (a, e)) in zip(actual, expected).enumerated() {
+            XCTAssertEqual(a, e, accuracy: accuracy, "Mismatch at index \(i)", file: file, line: line)
+        }
+    }
+
+    private func l2Norm(_ v: [Float]) -> Float {
```
a2they: nit: l2Norm function is unecessary, you can just do sqrt in the one place that calls this

Tests/SpeakerKitTests/SpeakerCentroidEmbeddingsTests.swift
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    // MARK: - Clusterer-level invariant: value matches post-reassignment mean
+
+    /// Regression for a2they's "correct centroid is properly set": for any embedding input
+    /// that drives VBxClustering end-to-end, the surfaced `speakerCentroids[k]` must equal
+    /// the arithmetic mean of the embeddings whose final `clusterIndices[i]` is `k`.
+    /// This holds irrespective of which internal path (VBx weighted, kMeans correction,
+    /// AHC fallback) produced the assignments, because the final recompute is always a
+    /// plain `centroidsFromAssignments(...)` over the post-reassignment labels.
```
a2they: You don't need to call me out :)

```suggestion
    /// For any embedding input that drives VBxClustering end-to-end, the surfaced
    /// `speakerCentroids[k]` must equal the arithmetic mean of the embeddings whose final
    /// `clusterIndices[i]` is `k`. This holds irrespective of which internal path
    /// (VBx weighted, kMeans correction, AHC fallback) produced the assignments, because the
    /// final recompute is always a plain `centroidsFromAssignments(...)` over the
    /// post-reassignment labels.
```

Tests/SpeakerKitTests/SpeakerCentroidEmbeddingsTests.swift
```diff
@@ -0,0 +1,379 @@
...
+        }
+
+        let clusterer = VBxClustering()
+        await clusterer.add(speakerEmbeddings: speakerEmbeddings)
+        let result = await clusterer.update(config: VBxClusteringConfig())
+
+        let clusters = result.clusterIndices
+        XCTAssertEqual(clusters.count, raw.count)
+        let kFinal = (clusters.max() ?? -1) + 1
+        XCTAssertGreaterThan(kFinal, 0, "expected at least one cluster")
```
a2they: pin kFinal to 2 so if default config changes in the future that collapses this we will revisit this test.
or you can validate clusters.max() with guard let else fail test so you dont need to do the ?? -1

```suggestion
        XCTAssertGreaterThan(kFinal, 2, "expected at least one cluster")
```

Tests/SpeakerKitTests/SpeakerCentroidEmbeddingsTests.swift
```diff
@@ -0,0 +1,379 @@
...
+        let result = try await speakerKit.diarize(audioArray: audioArray)
+
+        for (id, centroid) in result.speakerCentroidEmbeddings {
+            XCTAssertFalse(centroid.isEmpty, "centroid for speaker \(id) must not be empty")
+            for value in centroid {
+                XCTAssertTrue(value.isFinite, "centroid for speaker \(id) has non-finite value")
+            }
+            let norm = l2Norm(centroid)
+            XCTAssertGreaterThan(norm, 0.1, "centroid \(id) has suspiciously small norm \(norm)")
+            XCTAssertLessThan(norm, 10.0, "centroid \(id) has suspiciously large norm \(norm)")
```
a2they: We should probably use exact value or tigther range

Tests/SpeakerKitTests/DiarizationPipelinePerformanceTests.swift
```diff
@@ -0,0 +1,48 @@
...
+
+import XCTest
+import WhisperKit
+@testable import SpeakerKit
+
+/// End-to-end performance baseline for `speakerKit.diarize(...)` on the bundled VADAudio
+/// fixture. This file uses only pre-existing public API (`SpeakerKit`,
+/// `SpeakerKit.diarizer`, `Diarizer.loadModels`, `diarize(audioArray:)`) so it can be
+/// cherry-picked onto `main` to produce a baseline number against this branch.
+final class DiarizationPipelinePerformanceTests: XCTestCase {
```
a2they: this test is unrelated to centroid change. What does it achieve?

Sources/SpeakerKit/DiarizationResult.swift
```diff
@@ -101,6 +118,25 @@ public struct DiarizationResult: Sendable {
         self.segments = segments.sorted { $0.startFrame < $1.startFrame }
     }
 
+    // MARK: - Speaker Centroid Comparison
+
+    /// Cosine distance in `[0.0, 2.0]` between two speaker centroids from this result.
```
a2they, line 136: I would leave more clarification how to interpret the distances. What range is roughly the same speaker?

Sources/SpeakerKit/Pyannote/VBxClustering.swift
```diff
@@ -176,7 +199,7 @@ actor VBxClustering: Clusterer {
         return clusterIndices
     }
 
-    private func centroidsFromAssignments(assignments: [Int], embeddings: [[Float]], k: Int) -> [[Float]] {
+    func centroidsFromAssignments(assignments: [Int], embeddings: [[Float]], k: Int) -> [[Float]] {
```
a2they: update all call places intead of `k`
```suggestion
    func centroidsFromAssignments(assignments: [Int], embeddings: [[Float]], clusterCount: Int) -> [[Float]] {
```

Sources/SpeakerKit/Pyannote/VBxClustering.swift
a2they, line 92: nit: we should add a comment that these are used to seed cluster reassignment, the returned centroids are recomputed

Sources/SpeakerKit/Pyannote/VBxClustering.swift
```diff
@@ -22,15 +22,25 @@ actor VBxClustering: Clusterer {
...
 
-        let (clusters, _) = cluster(embeddings: _speakerEmbeddings, config: config)
+        let (clusters, _, centroids) = cluster(embeddings: _speakerEmbeddings, config: config)
 
         for (clusterIndex, clusterId) in clusters.enumerated() {
             _speakerEmbeddings[clusterIndex].clusterId = clusterId
         }
 
+        // expose centroids keyed by the final (post-reassignment) clusterId so downstream
+        // consumers see a centroid that matches the actual membership of each speakerId.
```
a2they:
```suggestion
        // Key centroids by the final clusterId so each entry matches its speakerId membership.
```

Sources/SpeakerKit/Pyannote/VBxClustering.swift
```diff
@@ -122,7 +132,20 @@ actor VBxClustering: Clusterer {
             }
         }
 
-        return (clusters, linkageMatrix)
+        // Recompute centroids from the final, post-reassignment cluster membership so the
+        // surfaced centroid for speaker k is the arithmetic mean of the embeddings currently
+        // labelled k across all three paths (VBx weighted, kMeans correction, AHC fallback).
```
a2they:
```suggestion
        // Returned centroids are the arithmetic mean of embeddings under the final assignment,
        // uniform across all paths (VBx weighted, kMeans correction, AHC fallback).
```

Sources/SpeakerKit/Pyannote/VBxClustering.swift
a2they: nit: move this outside (before) the if statement, so you don't need to re-do this on LOC 138

Sources/SpeakerKit/DiarizationResult.swift
a2they: you should included `speakerCentroidEmbeddings: [Int: [Float]] = [:]` in the public init as well

Sources/SpeakerKit/DiarizationResult.swift
```diff
@@ -101,6 +118,25 @@ public struct DiarizationResult: Sendable {
...
+    ///
+    /// Delegates to `MathOps.cosineDistance(_:_:)`, matching the convention used by
+    /// `MathOps.cosineDistanceMatrix` elsewhere in SpeakerKit. The result is clamped to
+    /// `[0, 2]` to absorb floating-point error near the extremes.
+    ///
+    /// - Returns: `nil` if either `speakerId` is absent from
+    ///   ``speakerCentroidEmbeddings``, the centroids have different dimensions, or either
+    ///   vector is empty. Zero-magnitude centroids (unreachable in real diarization runs)
+    ///   yield `MathOps.cosineDistance`'s sentinel of `1.0`.
+    public func centroidCosineDistance(between a: Int, _ b: Int) -> Float? {
```
a2they: do you have need for something like `public func nearestSpeakerCentroid(to embedding: [Float]) -> (speakerId: Int, distance: Float)?` ? can we be added later but I assume useful utility to find nearest speaker given embedding

Tests/SpeakerKitTests/DiarizationPipelinePerformanceTests.swift
leecrossley: Thanks, fair. It is orthogonal to the centroid API itself, so I dropped the file from this PR and kept the latency note in discussion only. If useful, happy to send the perf harness seperately against main since it only uses existing public API.

Sources/SpeakerKit/DiarizationResult.swift
leecrossley: Thanks. I don't think SpeakerKit publishes a cross-run centroid cutoff. The existing `clusterDistanceThreshold` is the within-run clustering knob, different layer. I tightened the docs to stable semantics only: distance in `[0, 2]`, `0` identical direction, `1` orthogonal, `2` opposite; zero-magnitude centroids return the existing `MathOps.cosineDistance` sentinel of `1.0`. Consumers pick same-speaker threshold per workflow.

Sources/SpeakerKit/DiarizationResult.swift
leecrossley: Thanks, added it. Overshow's shipping path only needs the exposed centroid map today, but this is usefull as the threshold-free lookup surface: `nearestSpeakerCentroid(to:)` returns the closest compatible centroid and leaves same-speaker policy to caller.

leecrossley:
Thanks again for the careful review - my project (Overshow) is a private desktop app, in the Swift helper we run SpeakerKit locally alongside WhisperKit. Per chunk: one diarise pass, each transcript segment gets the best timeoverlap speaker an corresponding centroid from `speakerCentroidEmbeddings` is persisted alongside the segment. Cosine distance on those centroids is used downstream (outside this helper) for cross chunk and cross session speaker reuse. We need to ensure that the returned centroid matches the final post-reassignment `speakerId`.

I pushed the last review pass as one batch: dropped the unrelated perf test from the diff, added public init parity, renamed `k` to `clusterCount`, hoisted the repeated embedding map, tightened the comments/tests, clarified distance semantics w/out inventing a same speaker cutoff and added `nearestSpeakerCentroid(to:)` as the threshold free lookup helper.

I also updated the PR body / resolved the addressed review threads after replying where useful, I'm happy to adjust anythign else / squash if helpful

Sources/SpeakerKit/DiarizationResult.swift
```diff
@@ -101,6 +128,49 @@ public struct DiarizationResult: Sendable {
...
+    /// `MathOps.cosineDistanceMatrix` elsewhere in SpeakerKit. The result is clamped to
+    /// `[0, 2]` to absorb floating-point error near the extremes. A distance of `0` means
+    /// identical direction, `1` means orthogonal vectors (no directional similarity), and
+    /// `2` means opposite direction.
+    ///
+    /// - Returns: `nil` if either `speakerId` is absent from
+    ///   ``speakerCentroidEmbeddings``, the centroids have different dimensions, or either
+    ///   vector is empty. Zero-magnitude centroids (unreachable in real diarization runs)
+    ///   yield `MathOps.cosineDistance`'s sentinel of `1.0`.
+    public func centroidCosineDistance(between a: Int, _ b: Int) -> Float? {
```
a2they: nit: the unlabeled second parameter reads awkwardly at call sites (i.e. `centroidCosineDistance(between: 0, 1)`). Suggest renaming to `between:and:` (public surface easier to fix now)
```suggestion
    public func centroidCosineDistance(between a: Int, and b: Int) -> Float? {
```

Sources/SpeakerKit/DiarizationResult.swift
```diff
@@ -101,6 +128,49 @@ public struct DiarizationResult: Sendable {
...
+
+    /// Nearest centroid in this result to an external speaker embedding.
+    ///
+    /// This is a pure nearest-neighbour lookup over ``speakerCentroidEmbeddings``. It does not
+    /// apply a same-speaker threshold; callers should interpret the returned distance according
+    /// to their own calibration.
+    ///
+    /// - Returns: The nearest compatible centroid, or `nil` when `embedding` is empty, no
+    ///   centroid exists, or all stored centroids have different dimensions.
+    public func nearestSpeakerCentroid(to embedding: [Float]) -> (speakerId: Int, distance: Float)? {
```
a2they, line 163: tie-breaking here depends on `[Int: [Float]]` iteration order, which isn't defined. Two centroids at the same distance can return a different `speakerId` between runs. For a cross-session matching helper this should be deterministic. Suggest iterating`speakerCentroidEmbeddings.keys.sorted()` and documenting that ties resolve to the lowest `speakerId`.

Sources/SpeakerKit/DiarizationResult.swift
```diff
@@ -31,26 +31,53 @@ public struct DiarizationResult: Sendable {
...
+    ///
+    /// Compare centroids with cosine distance via `centroidCosineDistance(between:_:)` or
+    /// `nearestSpeakerCentroid(to:)`, matching the convention used by
+    /// `MathOps.cosineDistanceMatrix` elsewhere in SpeakerKit. SpeakerKit does not define a
+    /// universal "same speaker" threshold for comparing centroids across independent runs;
+    /// callers should calibrate that policy for their model, audio, and application.
+    ///
+    /// This field is populated by the Pyannote backend (`PyannoteDiarizer`). Other backends
+    /// conforming to `Diarizer` may leave it as `[:]` if they do not expose per-cluster centroids.
+    public private(set) var speakerCentroidEmbeddings: [Int: [Float]]
```
a2they, line 53: One thing I wanted to flag on the centroid calculation, curious what you think.

The surfaced `speakerCentroidEmbeddings` is computed over all embeddings, but every centroid used internally in `VBxClustering.cluster(...)` is computed over the trainable subset only (the nonOverlappedFrameRatio > minActiveRatio filter). All three seed paths (VBx, kMeans, AHC fallback) use embeddingsFloats (trainable). Only the new surfaced centroid uses allEmbeddingsFloats.

So the centroid we return isn't quite the same kind of mean the pipeline itself uses. It folds in the overlap-flagged windows that the embedder is least confident on, which tends to pull the centroid toward the mixed-speaker region of the embedder's output space. Downstream consumers doing cosine matching end up with a noisier reference point than the one clustering already trusted.

Would it be worth adding an option on PyannoteDiarizationOptions, something like centroidSource: .finalAssignment | .trainableOnly, so callers can opt into the trainable-only centroid that matches the pipeline's internal convention? What was your testing like with trainable only vs all embeddings?

a2they (commented):
Thanks for addressing the previous PR feedback. Last few notes on the changes.

Sources/SpeakerKit/DiarizationResult.swift
leecrossley: good - added `PyannoteDiarizationOptions.centroidSource` with `.finalAssignment` as the current/default behaviour and `.trainableOnly` as the opt-in path

the trainable-only path still uses final postreassignment labels, but mean pools only embeddings that pass the clustering trainability filter. added a unit test pinning the difference between both sources

forr `.trainableOnly` if a final speaker has no trainable embeddings, I omit that speaker from `speakerCentroidEmbeddings` rather than surfacing a zero vector, whicg keeps `centroidCosineDistance(...)` returning nil for missing usable centroids instead of silently comparing against 1.0

Sources/SpeakerKit/DiarizationResult.swift
leecrossley: thanks - and agree, have renamed the public helper and updated the tests

Sources/SpeakerKit/DiarizationResult.swift
leecrossley: fixed so it now iterates sorted speaker ids, so equal-distance ties resolve deterministically to the lowest speakerId and added that to the doc comment and covered it in the test

Sources/SpeakerKit/DiarizationResult.swift
leecrossley: FYI we haven’'t run meaningful side-by-side multispeaker numbers yet, default is current behaviour

Sources/SpeakerKit/Pyannote/VBxClustering.swift
```diff
@@ -22,15 +22,28 @@ actor VBxClustering: Clusterer {
...
         _speakerEmbeddings.sort { ($0.windowIndex, $0.speakerIndex) < ($1.windowIndex, $1.speakerIndex) }
 
-        let (clusters, _) = cluster(embeddings: _speakerEmbeddings, config: config)
+        let (clusters, _, centroids) = cluster(embeddings: _speakerEmbeddings, config: config)
 
         for (clusterIndex, clusterId) in clusters.enumerated() {
             _speakerEmbeddings[clusterIndex].clusterId = clusterId
         }
 
+        // Key centroids by the final clusterId so each entry matches its speakerId membership.
```
a2they: `centroidsFromAssignments` happens to return an all-zero vector when a cluster has no members, and this loop reads "is the vector all zeros?" to decide whether to drop that key. That couples update() to a private quirk of the helper. If `centroidsFromAssignments` ever changes its empty representation, this filter silently breaks. It would also drop a real all-zero centroid if one ever showed up.

`centroidsFromAssignments` returning [[Float]] is right for its callers, but the new `centroidsFromFinalAssignments` has a different consumer: this loop, which builds a [Int: [Float]] map. If the helper returned that shape directly with empty clusters omitted, the post-filter and the clusterCount parameter both go away.

a cleaner change would be changing current:
```
    // helper signature: [[Float]] keyed positionally, empty clusters surface as zero vectors
    func centroidsFromFinalAssignments(
        assignments: [Int],
        embeddings: [SpeakerEmbedding],
        source: SpeakerCentroidSource,
        minActiveRatio: Float,
        clusterCount: Int
    ) -> [[Float]] { ... }

    // in cluster(): need numFinalClusters up front, conditional guards an empty result
    let numFinalClusters = (clusters.max() ?? -1) + 1
    let finalCentroids: [[Float]] = numFinalClusters > 0
        ? centroidsFromFinalAssignments(
            assignments: clusters,
            embeddings: embeddings,
            source: config.centroidSource,
            minActiveRatio: config.minActiveRatio,
            clusterCount: numFinalClusters)
        : []

    // in update(): post-filter loop looks for zero vector to drop
    for clusterId in Set(clusters.filter { $0 >= 0 }) where clusterId < centroids.count {
        guard centroids[clusterId].contains(where: { $0 != 0 }) else { continue }
        centroidMap[clusterId] = centroids[clusterId]
    }
```

to this flow:
```
    // helper signature: returns the map directly, empty clusters never get a key
    func centroidsFromFinalAssignments(
        assignments: [Int],
        embeddings: [SpeakerEmbedding],
        source: SpeakerCentroidSource,
        minActiveRatio: Float
    ) -> [Int: [Float]] {
        // single pass over embeddings:
        //     sums[assignment]   += embedding
        //     counts[assignment] += 1
        // skip embeddings rejected by the .trainableOnly filter.
        // divide each sum by its count.
    }

    // in cluster(): no numFinalClusters, no conditional
    let finalCentroids = centroidsFromFinalAssignments(
        assignments: clusters,
        embeddings: embeddings,
        source: config.centroidSource,
        minActiveRatio: config.minActiveRatio)

    // in update(): centroids is already the [Int: [Float]] map. the entire centroidMap
    // construction and the for-loop go away. forward it directly into the result.
    return ClusteringResult(
        clusterIndices: clusters,
        speakerEmbeddings: _speakerEmbeddings,
        speakerCentroids: centroids
    )
```

what do you think?

Sources/SpeakerKit/Pyannote/PyannoteConfig.swift
```diff
@@ -119,12 +119,25 @@ public class PyannoteConfig: SpeakerKitConfig, @unchecked Sendable {
 
 // MARK: - Diarization Options
 
+public enum SpeakerCentroidSource: Sendable {
```
a2they: nit: add `Equatable, Hashable`. auto-synthesized for case-only enums so callers/tests can do `centroidSource == .trainableOnly` instead of `if case .trainableOnly`

```suggestion
public enum SpeakerCentroidSource: Equatable, Hashable, Sendable {
```

Sources/SpeakerKit/Pyannote/PyannoteConfig.swift
```diff
@@ -119,12 +119,25 @@ public class PyannoteConfig: SpeakerKitConfig, @unchecked Sendable {
...
+    /// Mean of all embeddings under the final post-reassignment speaker labels.
```
a2they:
```suggestion
     /// Mean of all embeddings under the final post-reassignment speaker labels, without
    /// any filtering for quality of embeddings.
```

Sources/SpeakerKit/Pyannote/PyannoteConfig.swift
```diff
@@ -119,12 +119,25 @@ public class PyannoteConfig: SpeakerKitConfig, @unchecked Sendable {
...
+    case finalAssignment
+
+    /// Mean of the trainable subset under the final post-reassignment speaker labels.
+    ///
+    /// This excludes embeddings whose `nonOverlappedFrameRatio` does not pass the clustering
+    /// `minActiveRatio` filter, matching the subset used to seed VBx/k-means clustering.
+    /// Speakers with no trainable embeddings are omitted from the returned centroid map.
```
a2they:
```suggestion
    /// Mean of embeddings under the final post-reassignment speaker labels, with
    /// additional filtering by the clustering algorithm for purer voice embeddings.
    /// May omit speakers whose members are entirely excluded by the filter. Use
    /// `if let centroid = speakerCentroidEmbeddings[id]` rather than assuming the
    /// key is present.
```

Sources/SpeakerKit/DiarizationResult.swift
```diff
@@ -31,26 +31,55 @@ public struct DiarizationResult: Sendable {
...
 
+    /// Per-speaker centroid embeddings keyed by `speakerId`, in the raw speaker-embedder output
+    /// space (unnormalised, pre-PLDA). Useful for linking the same speaker across independent
+    /// `diarize(...)` calls without re-running the embedder.
+    ///
+    /// By default, each centroid is the arithmetic mean of the final per-window embeddings
+    /// assigned to that `speakerId` after clustering and cluster reassignment, so the centroid
+    /// reflects the speaker's actual membership in this result. Pyannote callers may set
+    /// `PyannoteDiarizationOptions.centroidSource` to `.trainableOnly` to mean-pool only the
+    /// trainable subset under those final assignments.
```
a2they: "the final per-window embeddings assigned to that speakerId" is ambiguous about which embeddings contribute. It is easier to defer to SpeakerCentroidSource docs than restate the same thing in two places.

I would repeat the caveat for .trainableOnly source in case someone is looking only at this docs.

```suggestion
    /// Each centroid is the mean of the per-window embeddings that ended up under that
    /// `speakerId`. Which embeddings contribute is controlled by
    /// `PyannoteDiarizationOptions.centroidSource`; see ``SpeakerCentroidSource``. Under
    /// `.trainableOnly`, some speakers in `segments` may not have a centroid in this
    /// map; use `if let centroid = speakerCentroidEmbeddings[id]` rather than assuming
    /// the key is present.
```

a2they (commented):
Looks great, thanks for addressing the feedback. I added some notes on docs and algorithm simplification. Please take a look.

Sources/SpeakerKit/Pyannote/VBxClustering.swift
leecrossley: good shout, the zero vector filter does couple to the helpers empty representation, have refactored

Tests/SpeakerKitTests/SpeakerCentroidEmbeddingsTests.swift
```diff
@@ -0,0 +1,474 @@
...
+            assignments: assignments,
+            embeddings: embeddings,
+            source: .trainableOnly,
+            minActiveRatio: 0.2
+        )
+        XCTAssertEqual(Set(trainableOnlyCentroids.keys), [0, 1])
+        assertVectorsEqual(trainableOnlyCentroids[0] ?? [], [1.0, 0.0])
+        assertVectorsEqual(trainableOnlyCentroids[1] ?? [], [0.0, 10.0])
+    }
+
```
a2they, line 251: The test above pins both clusters retain at least one trainable member. There's no test that exercises the "cluster has zero trainable members" path, which is the case the omitted-key contract specifically exists to handle.

I suggest adding a sibling test that explicitly asserts the cluster is absent from the map:

```suggestion
    /// `.trainableOnly` omits clusters whose members are all overlap-flagged, rather than surfacing a zero-vector centroid.
    func testCentroidsFromFinalAssignments_omitsClusterWithNoTrainableMembers() async {
        let embeddings = [
            SpeakerEmbedding(
                embedding: [1.0, 0.0],
                activeFrames: [1.0], windowIndex: 0, speakerIndex: 0,
                nonOverlappedFrameRatio: 1.0
            ),
            SpeakerEmbedding(
                embedding: [0.0, 1.0],
                activeFrames: [1.0], windowIndex: 1, speakerIndex: 0,
                nonOverlappedFrameRatio: 0.0
            ),
            SpeakerEmbedding(
                embedding: [0.0, 2.0],
                activeFrames: [1.0], windowIndex: 2, speakerIndex: 0,
                nonOverlappedFrameRatio: 0.0
            )
        ]
        let assignments = [0, 1, 1] // cluster 1 is entirely overlap-flagged

        let clusterer = VBxClustering()
        let trainable = await clusterer.centroidsFromFinalAssignments(
            assignments: assignments, embeddings: embeddings,
            source: .trainableOnly, minActiveRatio: 0.2
        )

        XCTAssertEqual(Set(trainable.keys), [0])
        assertVectorsEqual(trainable[0] ?? [], [1.0, 0.0])
    }
```

a2they (approved):
Thank you for the contribution and addressing the feedback! I suggested to add one more test. Otherwise PR looks good to merge!

a2they (approved):

a2they:
@leecrossley please take a look at the failing `SpeakerCentroidEmbeddingsTests` in https://github.com/argmaxinc/argmax-oss-swift/actions/runs/25588332570/job/75130102118?pr=463

leecrossley:
> @leecrossley please take a look at the failing `SpeakerCentroidEmbeddingsTests` in https://github.com/argmaxinc/argmax-oss-swift/actions/runs/25588332570/job/75130102118?pr=463

sorry yes thanks for flagging. the testCentroidsStableAcrossReruns was being too strict for an e2e CoreM backed diarisation rerun on the watchOS 26 sim. The exact centroid arithmetic tests still use `1e-5`, but this integration test now asserts that speaker ids are preserved and centroids stay within `1e-2` cosine distance. That covers the observed `~0.008` simulator drift while still being far tighter than the API same speaker matching contract

a2they (approved):
LGTM, thank you for the contribution!

PR #464, opened 2026-04-29
Reuse compatible iOS audio session in TTSKit playback

shura-v:
Adds a `preserveExistingAudioSession` option to `TTSKitConfig` for iOS playback.

By default, behavior is unchanged and `TTSKit` still configures `AVAudioSession` for `.playback` when starting playback. When `preserveExistingAudioSession` is enabled, `TTSKit` only activates the existing session and leaves its category, mode, and options untouched.

## Why

Some apps already own an active `.playAndRecord` session when starting TTS playback. Attempting to switch that shared session to `.playback` can fail at runtime, so this opt-in allows host apps to keep their existing audio session configuration.

## Testing

- `swift build --target TTSKit`
- Manual review of the iOS playback startup path

### DIFF
--- a/Sources/TTSKit/Qwen3TTS/Qwen3Config.swift
+++ b/Sources/TTSKit/Qwen3TTS/Qwen3Config.swift
@@ -235,6 +235,13 @@ open class TTSKitConfig {
     /// `nil` loads when `modelFolder` is non-nil, matching WhisperKit's default.
     public var load: Bool?
 
+    // MARK: - Playback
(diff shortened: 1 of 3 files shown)


a2they:
Hi @shura-v , thank you for the contribution. Please rebase your PR to pickup latest CI changes which should resolve the failing tests. We'll add these to next release once CI builds pass.

ZachNagengast (approved):

PR #465, opened 2026-04-29
Add playback callback

shura-v:
## Summary

This PR adds a playback-synchronized audio callback for TTS playback.

`TTSKit` already exposes generated audio through `SpeechCallback`, but that callback is tied to generation time, not actual playback time. For use cases like lip sync, waveform visualization, and other playback-reactive UI, consumers need audio samples that correspond to what is actually being scheduled/played by `AVAudioPlayerNode`.

## Why this is needed

Generation-time callbacks can drift from audible playback because playback depends on player scheduling, buffering, output timing, and suppression/test paths. For lip sync in particular, reacting to generated samples is not enough: the animation should follow what the user actually hears.

This PR introduces a callback at the playback layer so consumers can observe the audio that is scheduled for playback, rather than only the audio that was produced by the model.

## Notes from the previous PR discussion

This is a continuation/replacement of #445, which was accidentally closed after the source branch was deleted/restored.

A question came up there about whether this callback is intended to attach to what is actually scheduled in `playerNode`, so consumers receive samples as they are played/scheduled rather than as they are generated.

Yes — that is the intended distinction.

`SpeechCallback` remains useful for generation-time access to audio, but the new callback is meant for playback-synchronized consumers such as lip sync and visualizers.

Another question was whether output suppression effectively turns this into a custom playback backend API.

That is not the goal of this PR. `setOutputSuppressed(true)` is primarily useful for tests and non-audible playback paths. It makes the playback pipeline testable without producing sound, but this PR is not trying to promote suppression as a first-class replacement for the built-in audio playback system.

## Intended use cases

- Playback-synchronized lip sync
- Waveform / level visualization tied to actual playback
- Tests that verify playback callback behavior without audible output

## Non-goals

- Replacing `SpeechCallback`
- Providing a full custom audio backend API
- Making output suppression the primary way to consume generated TTS audio

## Related

Replaces / continues #445.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@
 
 [Argmax](https://argmaxinc.com/blog) Open-Source SDK Swift is a collection of turn-key on-device inference frameworks:
 - **WhisperKit** for speech-to-text with OpenAI Whisper
-- **SpeakerKit** for speaker diarization Pyannote
+- **SpeakerKit** for speaker diarization with Pyannote
 - **TTSKit** for text-to-speech with Qwen-TTS
 
 > [!IMPORTANT]
@@ -46,7 +46,7 @@
   - [Xcode Steps](#xcode-steps)
   - [Package.swift](#packageswift)
   - [Homebrew](#homebrew)
-- [Getting Started](#getting-started)
+- [WhisperKit](#whisperkit)
   - [Quick Example](#quick-example)
   - [Model Selection](#model-selection)
   - [Generating Models](#generating-models)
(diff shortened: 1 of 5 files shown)


a2they:
Hi @shura-v , thank you for the contribution. Please rebase your PR to pickup latest CI changes which should resolve the failing tests. We'll add these to next release once CI builds pass.

shura-v:
@a2they Rebased, thanks!

PR #473, opened 2026-05-12
Add tokenizerEndpoint to WhisperKitConfig (#388)

### LINKED ISSUE #388: The load tokenizer uses the default endpoint; hope adding the add `endpoint` parameter to download the tokenizer.
whisperkit.download allows downloading the CoreML model without using the default endpoint, but the Tokenizer download still uses the default endpoint.

https://github.com/argmaxinc/WhisperKit/blob/4ef384ec769b6fd51b49c0e8e74e8d8083e44926/Sources/WhisperKit/Core/WhisperKit.swift#L241-L252

https://github.com/argmaxinc/WhisperKit/blob/4ef384ec769b6fd51b49c0e8e74e8d8083e44926/Sources/WhisperKit/Utilities/ModelUtilities.swift#L20-L28

If I simply add the endpoint parameter to WhisperKitConfig and pass it down, the tokenizer should be able to be downloaded, right?

john-rocky:
Fixes #388.

Adds a `tokenizerEndpoint` option to `WhisperKitConfig` so tokenizer downloads honor a custom HuggingFace-compatible endpoint, mirroring `modelEndpoint`. When `tokenizerEndpoint` is `nil` it falls back to `modelEndpoint`, so the issue author's expected behavior — set one endpoint and both downloads route through it — works without extra wiring. Users that need separate endpoints (e.g. a private mirror that only hosts tokenizers) can still override.

### Changes
- `WhisperKitConfig.tokenizerEndpoint: String?` (optional, defaults to `nil`)
- `WhisperKit.tokenizerEndpoint` stored property, resolved at init as `config.tokenizerEndpoint ?? config.modelEndpoint`
- `ModelUtilities.loadTokenizer` gains optional `endpoint:` parameter, passed to `HubApiWrapper(endpoint:)`
- `UnitTests.testTokenizerEndpointFromConfig` covering 4 cases (default / fallback / override / tokenizer-only)

### Test
```
$ swift test --filter UnitTests/testTokenizerEndpointFromConfig
Test Case '-[WhisperKitTests.UnitTests testTokenizerEndpointFromConfig]' passed (0.002 seconds).
```
Adjacent `testTokenizerFolderFromConfig` still passes.

### Backward compatibility
All new parameters/properties are optional with `nil` defaults. The `convenience init` is left alone (parity with how `modelEndpoint` is exposed today).

Happy to broaden the scope to a `DownloadConfig` wrapper as you suggested in the issue thread — let me know if you'd rather have that landed first.

### DIFF
--- a/Sources/WhisperKit/Core/Configurations.swift
+++ b/Sources/WhisperKit/Core/Configurations.swift
@@ -15,6 +15,9 @@ open class WhisperKitConfig {
     public var modelToken: String?
     /// HuggingFace Hub compatible endpoint URL
     public var modelEndpoint: String?
+    /// HuggingFace Hub compatible endpoint URL for the tokenizer.
+    /// Falls back to `modelEndpoint` when `nil`. See #388.
+    public var tokenizerEndpoint: String?
     /// Folder to store models
     public var modelFolder: String?
     /// Folder to store tokenizers
@@ -77,6 +80,7 @@ open class WhisperKitConfig {
                 modelRepo: String? = nil,
                 modelToken: String? = nil,
                 modelEndpoint: String? = nil,
+                tokenizerEndpoint: String? = nil,
                 modelFolder: String? = nil,
                 tokenizerFolder: URL? = nil,
                 computeOptions: ModelComputeOptions? = nil,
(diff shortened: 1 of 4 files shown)


john-rocky:

PR #474, opened 2026-05-12
Add supportedLanguageCodes API for ModelVariant and WhisperKit (#398)

### LINKED ISSUE #398: Need an API to get all the available languages based on a model
Based on the model we download, I'm guessing the number of languages it supports will vary. Is there an API to detect what all languages a model supports? Or Is there a way to do it myself? Thanks in advance.

john-rocky:
Fixes #398. Adds a simple language-support introspection API on top of the existing variant detection so callers can decide which language list to show in their UI without parsing the model name themselves.

### Changes

- \`ModelVariant.supportedLanguageCodes: Set<String>\`
  - English-only (\`.tinyEn\` / \`.baseEn\` / \`.smallEn\` / \`.mediumEn\`) → \`["en"]\`
  - Multilingual up to \`large-v2\` → \`Constants.languageCodes\` minus \`yue\`
  - \`large-v3\` → \`Constants.languageCodes\` (the extra Cantonese token from [openai/whisper#1762](https://github.com/openai/whisper/discussions/1762))
- \`WhisperKit.supportedLanguageCodes\` forwards to the loaded variant.

The variant is detected from the loaded model via the existing \`ModelUtilities.detectVariant\` path, so this matches whatever \`modelVariant\` reports today — no new detection logic.

### Notes
- The turbo distillation (\`openai_whisper-large-v3-v20240930\`) shares the multilingual vocabulary of \`large-v3\` and is reported the same way here. As @ZachNagengast pointed out in the issue thread, non-English transcription quality on the turbo model is much weaker than the full \`large-v3\`, but that's a quality concern rather than a capability one. Happy to gate it differently if you'd prefer a separate variant case.
- I didn't touch \`Constants.languages\` (the name→code map) — its set of values is already the source of truth for \`Constants.languageCodes\`, which makes the variant-level filter trivial.

### Test
\`\`\`
$ swift test --filter UnitTests/testSupportedLanguageCodesByVariant
Test Case '-[WhisperKitTests.UnitTests testSupportedLanguageCodesByVariant]' passed (0.002 seconds).
\`\`\`
\`testSupportedLanguageCodesByVariant\` covers all 11 variants, the Cantonese inclusion on \`large-v3\`, and the \`WhisperKit\` forwarding (built with \`prewarm:false, load:false, download:false\` so it doesn't hit the network).

### DIFF
--- a/Sources/WhisperKit/Core/Models.swift
+++ b/Sources/WhisperKit/Core/Models.swift
@@ -59,6 +59,26 @@ public enum ModelVariant: CustomStringConvertible, CaseIterable {
         }
     }
 
+    /// Whisper language codes (ISO-639-1, or `yue` for Cantonese) supported by this variant.
+    ///
+    /// English-only variants (`.tinyEn`, `.baseEn`, `.smallEn`, `.mediumEn`) report
+    /// just `"en"`. Multilingual variants up to `large-v2` cover the original 99
+    /// Whisper languages. `large-v3` adds Cantonese (`yue`) on top of those.
+    ///
+    /// Note: the turbo distillation (`openai_whisper-large-v3-v20240930`) shares
+    /// the multilingual vocabulary of `large-v3` and is reported as such here, but
+    /// transcription quality outside English is not on par with the full model.
+    public var supportedLanguageCodes: Set<String> {
+        switch self {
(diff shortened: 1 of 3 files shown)


john-rocky:

PR #475, opened 2026-05-12
Scope MLMultiArray pixel-buffer locks via withUnsafeMutableBytes (#308)

### LINKED ISSUE #308: Usage of deprecated dataPointer or bytes properties warning in CoreML
I started noticing some model loading issues in version 15.3.x for older chips (M2). While looking at the CoreML logs, I noticed this warning. I don't think this is the cause, but I wanted to flag it in case anyone else sees this. 

```
2025-02-23 10:52:22.944382-0500 0xa52c89   Debug       0x0                  98093  0   (CoreML) [com.apple.coreml:coreml] Pixel buffer backing MLMultiArray is locked until this MultiArray is deallocated (or storage swapped) due to usage of deprecated dataPointer or bytes properties. Please use getBytesWithHandler instead.
2025-02-23 10:52:22.944392-0500 0xa52c89   Debug       0x0                  98093  0    (CoreML) [com.apple.coreml:coreml] Pixel buffer backing MLMultiArray is locked until this MultiArray is deallocated (or storage swapped) due to usage of deprecated dataPointer or bytes properties. Please use getBytesWithHandler instead.
```

From CoreML
```
    @available(macOS, introduced: 10.13, deprecated: 100000, message: "Use getBytesWithHandler or getMutableBytesWithHandler instead. For Swift, use withUnsafeBytes or withUnsafeMutableBytes.")
    open var dataPointer: UnsafeMutableRawPointer { get }
```

john-rocky:
Fixes #308.

Replaces \`MLMultiArray.dataPointer\` in WhisperKit/Core with scoped accessors (\`withUnsafeMutableBufferPointer\` / \`withUnsafeMutableBytes\`) so the pixel-buffer lock CoreML takes on storage access is released as soon as the closure returns, instead of being kept alive for the entire lifetime of the array. That's the exact deprecation the runtime log warns about:

> Pixel buffer backing MLMultiArray is locked until this MultiArray is deallocated (or storage swapped) due to usage of deprecated dataPointer or bytes properties. Please use getBytesWithHandler instead.

### Scope

Just the three places @ZachNagengast pointed at in the issue thread:

- \`Sources/WhisperKit/Core/Audio/AudioProcessor.swift\` — the \`MLMultiArray\` sample buffer that we hand back to the encoder is initialized inside \`withUnsafeMutableBufferPointer(ofType: Float.self)\`, so the source-copy + \`vDSP_vclr\` zero-padding share one scoped lock instead of leaving the buffer locked for the caller's lifetime.
- \`Sources/WhisperKit/Core/Text/LogitsFilter.swift\` — \`sumOfProbabilityOverTimestampsIsAboveAnyOtherToken\` wraps the BNNS log-softmax + log-sum-exp + max chain inside a single \`logits.withUnsafeMutableBytes\` so the locked region matches the actual BNNS work.
- \`Sources/WhisperKit/Core/Text/TokenSampler.swift\` — \`sampleWithBNNS\` does the same for the softmax/top-k/argmax path. The softmax output is its own allocation, so the array's bytes don't need to stay locked past the BNNS calls; \`makeArray\` reads from the descriptor, not from \`logits\`.

The deprecated \`BNNS.applyActivation\` warnings inside these functions are pre-existing and tracked in the inline TODOs — not touched here.

### TTSKit follow-up

TTSKit still uses \`dataPointer\` heavily in \`KVCache\`, \`PromptCache\`, \`EmbedTypes\`, \`Qwen3SpeechDecoder\`, \`Qwen3MultiCodeDecoder\`, and \`Sampling\`. Those are tighter loops where the locking pattern matters even more, but the rewrite is more invasive (per-step accessors and KV-cache snapshot/restore paths) so I kept this PR scoped to the original WhisperKit modules you flagged. Happy to send a follow-up once this one is in.

### Test

\`\`\`
$ swift build
Build complete!
$ swift test --filter "UnitTests/testSampler|UnitTests/testAudio|UnitTests/testLogitsFilter|UnitTests/testLogProb|UnitTests/testTimestamp"
Executed 13 tests, with 0 failures (0 unexpected) in 15.402 seconds.
\`\`\`

The full \`testInitTiny\` and \`testDecoderTokenizer\` paths (which exercise the sampler + audio processor end-to-end against a real \`tiny\` model) also pass locally on macOS / arm64.

### DIFF
--- a/Sources/WhisperKit/Core/Audio/AudioProcessor.swift
+++ b/Sources/WhisperKit/Core/Audio/AudioProcessor.swift
@@ -159,15 +159,27 @@ public extension AudioProcessing {
 
         let audioSamplesArray = try! MLMultiArray(shape: [NSNumber(value: frameLength)], dataType: .float32)
 
-        if actualFrameLength > 0 {
-            audioArray.withUnsafeBufferPointer { sourcePointer in
-                audioSamplesArray.dataPointer.assumingMemoryBound(to: Float.self).advanced(by: 0).initialize(from: sourcePointer.baseAddress!.advanced(by: startIndex), count: actualFrameLength)
+        // Use `withUnsafeMutableBufferPointer` so the CoreML pixel-buffer lock is
+        // scoped to this initialization rather than held until the MLMultiArray
+        // is deallocated (the deprecation warning behind issue #308).
+        audioSamplesArray.withUnsafeMutableBufferPointer(ofType: Float.self) { destPointer, _ in
+            if actualFrameLength > 0 {
+                audioArray.withUnsafeBufferPointer { sourcePointer in
+                    destPointer.baseAddress!.initialize(
+                        from: sourcePointer.baseAddress!.advanced(by: startIndex),
+                        count: actualFrameLength
+                    )
+                }
             }
-        }
-
-        // If the buffer is smaller than the desired frameLength, pad the rest with zeros using vDSP
-        if actualFrameLength < frameLength {
-            vDSP_vclr(audioSamplesArray.dataPointer.assumingMemoryBound(to: Float.self).advanced(by: actualFrameLength), 1, vDSP_Length(frameLength - actualFrameLength))
+            // If the buffer is smaller than the desired frameLength, pad the rest
+            // with zeros using vDSP.
+            if actualFrameLength < frameLength {
+                vDSP_vclr(
+                    destPointer.baseAddress!.advanced(by: actualFrameLength),
+                    1,
+                    vDSP_Length(frameLength - actualFrameLength)
+                )
+            }
         }
 
         return audioSamplesArray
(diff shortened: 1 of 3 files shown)


ZachNagengast:
@john-rocky Thanks for the contribution, this section of code uses BNNS which is deprecated and we'd prefer to remove it or replace it with pure Accelerate apis, which you are welcome to take on, but as is the BNNS code is considered fallback for older OS which wasn't the case when the original issue you referenced was opened.

Regarding the string of PRs you recently opened - I'm ok with the use of LLMs in generating PRs, and they do appear to be valid improvements for the codebase - however opening so many so quickly makes it appear like little care is going into each particular one, whereas the overhead to review is high considering the critical paths they are touching. This is a current trend on github (see openclaw) where people are sending LLMs out to resolve various open issues but end up having a counter effect on overall code quality and maintenance. Correct me if I'm wrong, but if that is the case, we'd greatly appreciate a specific error case that you encountered in your app, or even evidence that you could reproduce the error on your own system, versus spamming PRs in this way. Thats not to say the work isn't addressing a real issue, its more of a comment on the impression of low effort / validation.

john-rocky:
@ZachNagengast Thanks — you're right on both counts and I appreciate the honest framing.

These four PRs (#473, #474, #475, #476) were sweep fixes against open issues, not reproductions from my own app. I don't have the BNNS-fallback path triggered on a current OS and didn't independently verify the others end-to-end, so the validation overhead you're describing falls entirely on you. That's not the trade I should be asking you to make, especially when the touched paths are critical.

Closing all four. If I hit any of these for real in production, I'll reopen with a concrete repro (env, model, stack trace) rather than a generalized fix. Sorry for the noise.

PR #476, opened 2026-05-12
Lock WhisperKit.progress against torn-reference reads (#331)

### LINKED ISSUE #331: Data race detected by thread sanitizer on WhisperKit.progress instance variable.
The `progress` instance variable on WhisperKit is not being properly protected from data races.

Since WhisperKit is not actor-isolated, access to `progress` in the `runTranscribeTask` method can occur on multiple threads.

When I run in my own app and enable the thread sanitizer in the scheme diagnostic options, it detects this section as a data race:

```
if progress.isFinished {
   // Reset progress if it is completed
   progress = Progress()
}
```

Here's the detailed thread sanitizer report:
```
WARNING: ThreadSanitizer: data race (pid=51838)
  Read of size 8 at 0x000106e68c58 by thread T29:
    #0 WhisperKit.WhisperKit.progress.getter : __C.NSProgress <null> (TranscrybeCore:arm64+0x2c5428)
    #1 (5) suspend resume partial function for WhisperKit.WhisperKit.runTranscribeTask(audioArray: Swift.Array<Swift.Float>, decodeOptions: Swift.Optional<WhisperKit.DecodingOptions>, callback: Swift.Optional<(WhisperKit.TranscriptionProgress) -> Swift.Optional<Swift.Bool>>) async throws -> Swift.Array<WhisperKit.TranscriptionResult> <null> (TranscrybeCore:arm64+0x2fac20)
    #2 swift::runJobInEstablishedExecutorContext(swift::Job*) <null> (libswift_Concurrency.dylib:arm64e+0x5c450)

  Previous write of size 8 at 0x000106e68c58 by thread T35:
    #0 (5) suspend resume partial function for WhisperKit.WhisperKit.runTranscribeTask(audioArray: Swift.Array<Swift.Float>, decodeOptions: Swift.Optional<WhisperKit.DecodingOptions>, callback: Swift.Optional<(WhisperKit.TranscriptionProgress) -> Swift.Optional<Swift.Bool>>) async throws -> Swift.Array<WhisperKit.TranscriptionResult> <null> (TranscrybeCore:arm64+0x2fad34)
    #1 swift::runJobInEstablishedExecutorContext(swift::Job*) <null> (libswift_Concurrency.dylib:arm64e+0x5c450)

  Location is heap block of size 712 at 0x000106e68a00 allocated by main thread:
    #0 malloc <null> (libclang_rt.tsan_osx_dynamic.dylib:arm64e+0x55fb4)
    #1 _malloc_type_malloc_outlined <null> (libsystem_malloc.dylib:arm64e+0x1da80)
    #2 swift::runJobInEstablishedExecutorContext(swift::Job*) <null> (libswift_Concurrency.dylib:arm64e+0x5c450)
    #3 __debug_main_executable_dylib_entry_point <null> (Transcrybe.debug.dylib:arm64+0x7ed0)
    #4 <null> <null> (0x000193a1ab4c)

  Thread T29 (tid=6180198, running) is a GCD worker thread

  Thread T35 (tid=6180204, running) is a GCD worker thread

SUMMARY: ThreadSanitizer: data race (/Users/chris/Library/Developer/Xcode/DerivedData/Transcrybe-hbenxjemepmsovdjjuucnlyerfnc/Build/Products/Debug/Transcrybe.app/Contents/Frameworks/TranscrybeCore.framework/Versions/A/TranscrybeCore:arm64+0x2c5428) in WhisperKit.WhisperKit.progress.getter : __C.NSProgress+0x38
```

Image

john-rocky:
Fixes #331.

ThreadSanitizer flags a data race between callers reading \`progress.isFinished\` / \`fractionCompleted\` from one thread (typical for KVO/Combine subscribers updating a UI) and the \`progress = Progress()\` writes inside \`runTranscribeTask\`. Replacing the reference is one pointer-sized store on its own, but if a reader is fetching that storage at the same instant it can observe a torn value, and the report in the issue lines up with that exact split between the getter and the assignment.

The fix mirrors the locking pattern already used elsewhere in ArgmaxCore (\`UnfairLock\` from \`ConcurrencyUtilities.swift\`):

- back \`progress\` with a private \`_progress\` field plus a private \`progressLock = UnfairLock()\`;
- expose \`progress\` as a computed property that returns \`_progress\` under the lock — KVO/Combine subscribers (e.g. \`testVADProgress\`) keep their existing access pattern;
- wrap the two reset paths (\`if _progress.isFinished\` after a successful run and the cancellation \`catch\` arm) in a single locked compare-and-set so concurrent readers either see the old \`Progress\` or the freshly allocated one — never a half-replaced pointer.

This keeps \`WhisperKit\` non-actor-isolated, which preserves the current public API while @ZachNagengast's planned strict-concurrency work lands later.

### Test
\`\`\`
$ swift build
Build complete!
$ swift test --filter UnitTests/testInit
Executed 2 tests, with 0 failures (0 unexpected) in 4.396 seconds.
\`\`\`

The reporter's repro (live transcription loop with TSan enabled) reads progress on the main actor and writes it from the transcribe task, which the new lock now serializes. I didn't add a unit-level concurrency test because the racing write only happens from \`runTranscribeTask\`, which requires a loaded model and a real audio buffer — \`testVADProgress\` already exercises the read path end-to-end.

### Notes
- \`UnfairLock\` is non-reentrant, but neither write call path acquires the lock recursively (\`runTranscribeTask\` is the only writer and only enters once per outcome).
- The \`Sendable\` story for \`WhisperKit\` as a whole is still open (per the issue thread), so this PR intentionally doesn't add a \`@unchecked Sendable\` annotation alongside it.

### DIFF
--- a/Sources/WhisperKit/Core/WhisperKit.swift
+++ b/Sources/WhisperKit/Core/WhisperKit.swift
@@ -41,7 +41,15 @@ open class WhisperKit {
 
     /// Progress
     public private(set) var currentTimings: TranscriptionTimings
-    public private(set) var progress = Progress()
+    /// Reported transcription progress. Reads go through `progressLock` so the
+    /// reference can be safely reset from inside `runTranscribeTask` while
+    /// callers are observing it from another thread (see issue #331).
+    public var progress: Progress {
+        progressLock.withLock { _progress }
+    }
+
+    private var _progress = Progress()
+    private let progressLock = UnfairLock()
 
     /// Configuration
     public var modelFolder: URL?
@@ -1002,17 +1010,21 @@ open class WhisperKit {
                 transcribeTaskResult.logTimings()
             }
 
-            if progress.isFinished {
-                // Reset progress if it is completed
-                progress = Progress()
+            // Atomically reset when finished so callers reading
+            // `progress.isFinished` from another thread can't observe a half-
+            // replaced reference.
+            progressLock.withLock {
+                if _progress.isFinished {
+                    _progress = Progress()
+                }
             }
 
             return [transcribeTaskResult]
         } catch {
             // Handle cancellation
             if error is CancellationError {
-                // Reset progress when cancelled
-                progress = Progress()
+                // Reset progress when cancelled (same data-race story as above).
+                progressLock.withLock { _progress = Progress() }
             }
             throw error
         }


john-rocky:

PR #481, opened 2026-05-18
TTSKit: guard degenerate distributions in token sampling

yemreak:
### Summary
Three sites in `Sampling.swift` draw `Float.random(in: 0..<probSum)` over a sampled distribution. When that sum is zero (or negative) the call traps (`0..<0` is an invalid range), or yields an undefined index further down. Replace the random draw with an argmax fallback over the same distribution so the generator stays deterministic instead of crashing.

### Motivation
Encountered with rare numerical edge cases during Qwen3 TTS generation in long contexts (the resampled logits collapse to all-zero after the top-k mask). The crash is hard to diagnose because the trace points inside `Foundation`, not the model code.

### Changes
* Top-k branch — fall back to `argmax(probs)` when `probSum <= 0`.
* Top-p / plain branch — same guard before the random draw.
* In-place `vDSP_vsdiv` normalization — fall back to `vDSP_maxvi` when the running sum is non-positive.

### Testing
Black-box: regenerated the offending Qwen3 prompt and confirmed the crash no longer occurs and output stays sensible. No behavioural change on healthy distributions.

### DIFF
--- a/Sources/ArgmaxCLI/ArgmaxCLI.swift
+++ b/Sources/ArgmaxCLI/ArgmaxCLI.swift
@@ -4,7 +4,7 @@
 import ArgumentParser
 import Foundation
 
-let VERSION: String = "development"
+let VERSION: String = "v1.0.0"
 
 var subcommands: [ParsableCommand.Type] {
 #if BUILD_SERVER_CLI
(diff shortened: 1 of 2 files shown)


Sources/ArgmaxCLI/ArgmaxCLI.swift
```diff
@@ -4,7 +4,7 @@
 import ArgumentParser
 import Foundation
 
-let VERSION: String = "development"
+let VERSION: String = "v1.0.0"
```
ZachNagengast, line 7:
```suggestion
let VERSION: String = "development"
```
This probably came in when branching from the release/v1.0.0 branch

PR #487, opened 2026-06-09
Fix crash in GreedyTokenSampler when top-K probability sum is zero

### LINKED ISSUE #478: Fatal crash in GreedyTokenSampler.sampleFromProbs when top-K probability sum is zero
## Summary

`GreedyTokenSampler.sampleFromProbs` crashes with `Fatal error: Can't get random value with an empty range` when the sum of top-K probabilities is zero.

## Location

`Sources/TTSKit/Utilities/Sampling.swift`, line 239 (v1.0.0):

```swift
let probSum = probsArray.reduce(0, +)
let randomValue = Float.random(in: 0..<probSum, using: &rng)  // crashes when probSum == 0``

## Cause
When softmax outputs become numerically unstable (e.g., all logits are -infinity or extremely small), the top-K probabilities all round to 0. This makes prob​Sum = 0, resulting in Float​.random(in: 0..<0) — an empty range that triggers a Swift runtime fatal​Error.

## Suggested Fix
Guard against prob​Sum <= 0 and fall back to the highest-probability token (argmax):

```swift
guard probSum > 0 else {
    let maxIdx = probsArray.enumerated().max(by: { $0.element < $1.element })?.offset ?? 0
    return Int32(idxArray[maxIdx])
}
let randomValue = Float.random(in: 0..<probSum, using: &rng)``

## Environment
• TTSKit v1.0.0
• iOS 18+
• Observed during on-device TTS generation

### PR DESCRIPTION
**Pull request by tsushanth**
## Summary

- Guard `Float.random(in: 0..<probSum)` in `GreedyTokenSampler.sampleFromProbs` against degenerate softmax output.
- Fall back to argmax when top-K probabilities sum to zero or NaN.
- Add a regression test that reproduces the original crash.

## Problem

When softmax output becomes numerically degenerate — all logits at `-inf`, NaN propagation, or extreme underflow — the top-K probabilities sum to zero. `Float.random(in: 0..<probSum)` then traps with:

```
Fatal error: Can't get random value with an empty range
```

I confirmed this locally with the new test: temporarily reverting the guard reproduces the exact crash from Swift's `FloatingPointRandom.swift:52`.

## Fix

Inline guard in [`Sources/TTSKit/Utilities/Sampling.swift`](https://github.com/argmaxinc/WhisperKit/blob/main/Sources/TTSKit/Utilities/Sampling.swift). When `probSum` is zero or non-finite, return the highest-probability token from the top-K window (argmax fallback). Lock-free, no new dependencies, no API changes.

## Test plan

- [x] New regression test `testGreedySamplerZeroProbSumFallsBackToArgmax` constructs `[1, 1, vocabSize]` logits at `-Float.infinity`, calls `sampleCodec0` with `topK=4`, asserts no crash and a valid token index.
- [x] Verified the test reproduces the original crash without the guard.
- [x] All 5 existing `GreedyTokenSampler` tests pass.
- [x] `swift build` succeeds.

Fixes #478

### DIFF
--- a/Sources/TTSKit/Utilities/Sampling.swift
+++ b/Sources/TTSKit/Utilities/Sampling.swift
@@ -236,6 +236,12 @@ public class GreedyTokenSampler: TokenSampling, @unchecked Sendable {
             let probsArray = await topKProbs.toFloatArray()
             let idxArray = await topKIndices.toIntArray()
             let probSum = probsArray.reduce(0, +)
+            // Numerically degenerate softmax (all-zero or NaN probs) would make
+            // `Float.random(in: 0..<probSum)` trap. Fall back to argmax instead.
+            guard probSum.isFinite, probSum > 0 else {
+                let maxIdx = probsArray.enumerated().max(by: { $0.element < $1.element })?.offset ?? 0
+                return Int32(idxArray[maxIdx])
+            }
             let randomValue = Float.random(in: 0..<probSum, using: &rng)
             var cumulativeSum: Float = 0
             for (i, probability) in probsArray.enumerated() {
--- a/Tests/TTSKitTests/TTSKitUnitTests.swift
+++ b/Tests/TTSKitTests/TTSKitUnitTests.swift
@@ -291,6 +291,30 @@ final class TTSKitUnitTests: XCTestCase {
         XCTAssertEqual(token, 5, "Suppressed token should be skipped, picking next best")
     }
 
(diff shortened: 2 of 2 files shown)


tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

PR #506, opened 2026-07-13
TTSKit: fused MultiCodeDecoder — one CoreML call per Qwen3 RVQ frame

mjfrey:
## What

Each Qwen3-TTS talker frame currently expands into 15 residual codes by running the code_predictor **16 times** (prefill ×2, then one prediction per code), with a host-side sampling pass and a codec-embedding lookup between every call. Per frame that is 16 CoreML dispatches + 15 samplings + 15 embedding lookups — the dominant cost of the generation loop.

This PR adds runtime support for a **fused** MultiCodeDecoder variant: the entire frame as **one CoreML program**. The 16 positions are unrolled in-graph (the inner context is a static 16, the head count a static 15), each next-step embedding is gathered from the codec embedding table in-graph, and sampling happens in-graph via the **Gumbel-max trick**: the runtime supplies pre-drawn Gumbel(0,1) noise and the graph computes `argmax(logits/T + G)` over the top-k-masked logits — mathematically the same distribution as the stepped host sampler, and deterministic given the noise (which is what makes the equivalence directly testable). The model returns the 15 codes plus the summed code embeddings, so the runtime just adds `CodeEmbedder(code0)` and the text embed to form the next talker input.

`scripts/fused_multi_code_decoder/` contains the conversion script (checkpoint-agnostic: Base and CustomVoice share the `talker.code_predictor` tensor layout), the PyTorch reference modules it validates against, and a README.

## Behavior

- The fused variant is detected at load time by its `gumbel` input. **All existing stepped models are untouched and remain the default path** — with no fused asset present this PR is behaviorally invisible.
- `temperature` is a runtime input; **top-k is baked into the graph at conversion time** (`torch.topk` needs a static k to trace) via the script's `--top-k`.
- Prewarm branches to full-frame dummy predictions for fused models.
- One caveat: fused-path noise currently comes from the system RNG, so `GenerationOptions.seed` does not reproduce fused audio (the stepped path keeps that property). Happy to thread the seeded sampler through if you'd prefer — see open questions.

## Correctness

- The fused torch graph reproduces a stepped PyTorch reference **exactly** (identical codes, bit-exact embedding sum) under identical noise; the conversion script gates on this before converting.
- Fused fp16 CoreML vs stepped fp16 CoreML, identical noise, 20 random frames: **99.3% code agreement (298/300), 19/20 frames bit-identical**, mean first divergence at position 14.8 of 15 — i.e. the rare disagreement is a last-code sampling tie, not drift. (Comparing fp16 against the fp32 reference instead shows ~65% — autoregressive tie-flips cascade, so that number overstates the difference; the stepped fp16 model diverges from fp32 the same way.)

## Performance (Apple silicon Mac, 0.6B, W8A16, same text/seed/settings)

| Metric | Stepped | Fused | Δ |
|---|---|---|---|
| End-to-end generation | 19.8 steps/s | 24.3 steps/s | **+23%** |
| Code-predictor frame (prediction only) | 62.9 ms (16 calls) | 43.8 ms (1 call) | **1.44×** |
| Host work per frame | 15 samplings + 15 embedding lookups | none | eliminated |

The end-to-end number is the meaningful one: on phone-class devices, where per-dispatch overhead is proportionally larger, the effect is at least as big — validated in a production app on an A19 iPhone and an M5 iPad, where the added headroom made 1.25× time-stretched playback (which needs ~17% above realtime) run cleanly.

Compiled fused asset (8-bit palettized, embedding table included): 137 MB, deployed as `multi_code_decoder/<versionDir>/<variant>/MultiCodeDecoder.mlmodelc` and selected via `TTSKitConfig.multiCodeDecoderVariant`.

## Open questions

1. **Baked top-k**: our asset bakes 20 to match our runtime settings; your default is 50. Bake 50 for shipped assets, or would you rather explore top-k as a graph input?
2. **Seeded noise**: should the fused path draw its Gumbel noise through the seeded `GreedyTokenSampler` RNG to preserve `seed` reproducibility? Straightforward to add if wanted.
3. **Assets**: shipping fused variants for the CustomVoice models is your side (weights differ from the checkpoint we converted); the script in `scripts/fused_multi_code_decoder/` is the complete recipe and gates itself against the PyTorch reference.

`swift build` and TTSKit unit tests (85) pass; validated in continuous on-device streaming use.


EduardoPach:
Thanks, @mjfrey! This was indeed a meaningful optimization, an amazing contribution. We've merged your proposal into https://github.com/argmaxinc/argmax-oss-swift/pull/513 with some small changes to how `MultiCodeDecoder` is packaged, therefore closing this PR

PR #511, opened 2026-07-26
fix: normalize NLLanguage code so Chinese hits the no-space word split path

### LINKED ISSUE #510: Chinese word timestamps broken: splitToWordTokens language gate never matches "zh" (NLLanguage returns "zh-Hans"/"zh-Hant")
## Summary

With `wordTimestamps: true`, Chinese transcriptions produce sentence-length "words" with severely wrong timings. Root cause: the no-space-language gate in `splitToWordTokens` compares `NLLanguageRecognizer` output against Whisper-style codes, but `NLLanguage` raw values for Chinese are BCP-47 tags (`"zh-Hans"` / `"zh-Hant"`), which never equal `"zh"`. Chinese therefore falls through to the space-based splitter designed for Latin languages.

## Reproduction

Transcribe any Mandarin audio with `DecodingOptions(language: "zh", wordTimestamps: true)` and inspect `allWords`:

```
splitToWordTokens actual:  ["<|0.00|>今天天气很好", "，我们来做个测试吧", "？", "<|endoftext|>"]
expected (per-token, like ja/th): ["<|0.00|>", "今天", "天", "气", "很好", "，", "我们", "来", "做", "个", "测", "试", "吧", "？", "<|endoftext|>"]
```

## Root cause

`Sources/WhisperKit/Core/Models.swift`, `splitToWordTokens`:

```swift
let languageCode = recognizer.dominantLanguage?.rawValue
if ["zh", "ja", "th", "lo", "my", "yue"].contains(languageCode) { ... }
```

`NLLanguageRecognizer` probe results against that whitelist:

| Input text | `dominantLanguage.rawValue` | Matches whitelist |
|---|---|---|
| Simplified Chinese | `zh-Hans` | ❌ |
| Traditional Chinese | `zh-Hant` | ❌ |
| Japanese | `ja` | ✅ |
| Thai | `th` | ✅ |
| Lao | `lo` | ✅ |
| Burmese | `my` | ✅ |

Every whitelisted language matches except both Chinese variants. (`"yue"` is also unreachable — `NLLanguage` has no Cantonese value; Cantonese text is detected as `zh-Hant`.)

For reference, openai/whisper checks its own decoding language code (literally `"zh"`) rather than re-detecting from text ([tokenizer.py#L277-L282](https://github.com/openai/whisper/blob/main/whisper/tokenizer.py)), so this is a porting regression.

## Impact

The DTW alignment itself is computed per token and is fine; the broken grouping merges whole clauses into single "words", whose durations then get rewritten by the max-word-duration heuristic (`min(0.7, median) * 2` = 1.4s cap in `SegmentSeeker`). On a real 209.9s Mandarin monologue (large-v3):

| Metric | Current | After per-token split |
|---|---|---|
| Word count | 69 (avg 8.8 chars, max 24) | 481 (avg 1.29 chars) |
| Median word duration | 1400ms (the clamp value) | 220ms |
| Words with duration == 1.4s | 27 | 0 |
| Timeline coverage by words | 45% | 60% (consistent with VAD silence measurement) |

Anything built on Chinese word timings (subtitle karaoke, transcript-based editing, silence trimming) currently cuts into real speech.

## Suggested fix

Normalize the detected language through `Locale` before the whitelist check so `zh-Hans`/`zh-Hant` map to `zh` (also transparently fixes Cantonese via the `zh` path). PR with a one-line fix plus a `testSplitToWordTokensChinese` regression test (mirroring the existing Japanese test; fails before, passes after): #511

A deeper alternative would be passing the decode language into the tokenizer as openai/whisper does, but that changes the `WhisperTokenizer` protocol surface — left to maintainers' judgment.

freecodetiger:
Fixes #510

## Problem

`splitToWordTokens` gates the no-space-language split on `NLLanguageRecognizer` output, but `NLLanguage` raw values for Chinese are BCP-47 tags (`"zh-Hans"` / `"zh-Hant"`) which never equal the whitelisted `"zh"`. Chinese text falls through to the space-based splitter, gluing whole clauses into single "words":

```
before: ["<|0.00|>今天天气很好", "，我们来做个测试吧", "？", "<|endoftext|>"]
after:  ["<|0.00|>", "今天", "天", "气", "很好", "，", "我们", "来", "做", "个", "测", "试", "吧", "？", "<|endoftext|>"]
```

Downstream, these sentence-length "words" hit the max-word-duration clamp (`min(0.7, median) * 2` = 1.4s) in `SegmentSeeker`, producing severely wrong word timestamps for zh-Hans, zh-Hant, and Cantonese (detected as zh-Hant). The per-token DTW alignment itself is correct — only the grouping is wrong — so this change strictly improves timing granularity with no alignment changes.

## Change

- `Models.swift`: normalize the detected language through `Locale` before the whitelist check (`zh-Hans`/`zh-Hant` → `zh`). Matches openai/whisper, which checks its decoding language code directly ([tokenizer.py#L277-L282](https://github.com/openai/whisper/blob/main/whisper/tokenizer.py)). No behavior change for other languages: `ja`/`th`/`lo`/`my`/`en` etc. map to themselves.
- `UnitTests.swift`: add `testSplitToWordTokensChinese`, mirroring `testSplitToWordTokensJapanese`. Fails before this change, passes after.

## Validation

- New Chinese test: red on `main`, green with this change.
- Existing `testSplitToWordTokens` (en) and `testSplitToWordTokensJapanese` expectations verified unchanged.
- Real-footage check (209.9s Mandarin monologue, large-v3, `wordTimestamps: true`):

| Metric | Before | After |
|---|---|---|
| Word count | 69 (avg 8.8 chars, max 24) | 481 (avg 1.29 chars) |
| Median word duration | 1400ms (clamp artifact) | 220ms |
| Words clamped at exactly 1.4s | 27 | 0 |
| Timeline coverage | 45% | 60%, consistent with independent energy-VAD silence measurement (~98% agreement on pause locations) |

## Notes

- `"yue"` in the whitelist remains unreachable (`NLLanguage` has no Cantonese; such text is detected as `zh-Hant` and now correctly takes the `zh` path).
- A design-level alternative is threading the decode language into the tokenizer as openai/whisper does; that touches the `WhisperTokenizer` protocol, so it is out of scope here.

### DIFF
--- a/Tests/WhisperKitTests/UnitTests.swift
+++ b/Tests/WhisperKitTests/UnitTests.swift
@@ -1393,6 +1393,29 @@ final class UnitTests: XCTestCase {
         XCTAssertEqual(wordTokens, expectedWordTokens, "Word tokens did not match expected output in Unicode split.")
     }
 
+    func testSplitToWordTokensChineseTraditional() async throws {
+        let tokenizer = try await ModelUtilities.loadTokenizer(for: .tiny)
+
+        // 今天天氣很好，我們來做個測試吧？
+        let tokenIds = [50364, 12074, 6135, 24090, 23801, 171, 120, 234, 5884, 3763, 10907, 3338, 9592, 105, 22099, 6062, 171, 120, 253, 50257]
+        let originalWords = tokenIds.map { tokenizer.convertIdToToken($0) }
+
+        let (words, wordTokens) = tokenizer.splitToWordTokens(tokenIds: tokenIds)
+
+        // Traditional Chinese (zh-Hant) must also normalize to "zh" to hit the no-space-language path.
+        // First, conservative guard assertions that don't depend on specific CFStringTokenizer output:
+        XCTAssertNotEqual(originalWords, words, "Traditional Chinese must hit the no-space split path, not fall back to 1:1 token mapping")
+        let flattenedTokens = wordTokens.flatMap { $0 }
+        XCTAssertEqual(tokenIds, flattenedTokens, "Split must preserve all input tokens without reordering")
+
+        // Then, exact match against known-good split output (same pattern as Simplified Chinese test):
+        let expectedWords = ["<|0.00|>", "今天", "天", "氣", "很好", "，", "我們", "來", "做", "個", "測", "試", "吧", "？", "<|endoftext|>"]
+        let expectedWordTokens = [[50364], [12074], [6135], [24090], [23801], [171, 120, 234], [5884], [3763], [10907], [3338], [9592, 105], [22099], [6062], [171, 120, 253], [50257]]
+
+        XCTAssertEqual(words, expectedWords, "Words did not match expected output in Unicode split.")
+        XCTAssertEqual(wordTokens, expectedWordTokens, "Word tokens did not match expected output in Unicode split.")
+    }
+
     // MARK: - Options Tests
 
     func testSampleLength() async throws {
--- a/Tests/WhisperKitTests/UnitTests.swift
+++ b/Tests/WhisperKitTests/UnitTests.swift
@@ -1374,6 +1374,25 @@ final class UnitTests: XCTestCase {
         XCTAssertEqual(wordTokens, expectedWordTokens, "Word tokens did not match expected output in Unicode split.")
     }
 
+    func testSplitToWordTokensChinese() async throws {
+        let tokenizer = try await ModelUtilities.loadTokenizer(for: .tiny)
+
+        // 今天天气很好，我们来做个测试吧？
+        let tokenIds = [50364, 12074, 6135, 42204, 23801, 171, 120, 234, 15003, 6912, 10907, 7549, 11038, 233, 5233, 243, 6062, 171, 120, 253, 50257]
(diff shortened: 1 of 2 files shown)


a2they:
Can you please add a unit test for Traditional Chinese / zh-Hant?

Tests/WhisperKitTests/UnitTests.swift
```diff
@@ -1374,6 +1374,25 @@ final class UnitTests: XCTestCase {
         XCTAssertEqual(wordTokens, expectedWordTokens, "Word tokens did not match expected output in Unicode split.")
     }
 
+    func testSplitToWordTokensChinese() async throws {
```
a2they, line 1377: Can you also add a unit test for Traditional Chinese / zh-Hant?

a2they (approved):
LGTM, thank you for the contribution. One simple request to add UT for zh-Hant.

Tests/WhisperKitTests/UnitTests.swift
freecodetiger: Thanks for the review! Added testSplitToWordTokensChineseTraditional for zh-Hant coverage — CI is green. PTAL.

PR #520, opened 2026-08-05
Support the legacy single-function SpeechDecoder and MultiCodeDecoder assets

EduardoPach:
#494 and #513 moved the SpeechDecoder and the MultiCodeDecoder onto multifunction
assets, which dropped support for the legacy single-function `W8A16` ones. This
restores both by detecting the layout at load time instead of assuming it.

### SpeechDecoder

- Probe `MLModelAsset.functionNames`; empty → load without a `functionName`.
  Requesting `throughputOptimized` against it errors instead of silently
  producing latency-mode audio.
- Read the `kv_cache_update_mask` rank from the model (2 = legacy, 3 =
  multifunction). Not derivable from `codesPerStep` — both ranks exist at
  `codesPerStep == 1`, so `SpeechDecoderCache` takes it as a parameter.
- `ModelUtilities` dimension reads are now bounds-safe; probing an
  unexpected-rank input used to trap.

### MultiCodeDecoder

Same probe: the legacy asset is schema-identical to the multifunction `stepped`
graph, so it loads without a `functionName` and drives the same decode path.
`.fused` has no counterpart in a single-function asset, so asking for it reports
a configuration error naming the multifunction variant rather than quietly
running stepped. `.stepped` stays the default; both variants are supported.

### Downloader

`setupModels` built a fresh `TTSKitConfig` for the download that dropped
`versionDir` and the six component variants. `downloadPatterns` is derived from
those, so any pinned variant was ignored — the downloader fetched the preset
defaults and the load then failed on the variant actually requested.

This is pre-existing (it reproduces on `main`), but it makes the legacy support
above unreachable through `TTSKit(config:)`, so it is fixed here. It also means a
user who already has `W8A16` and pins it currently pulls the ~1.1 GB
multifunction assets they don't need before failing to load; after the fix that
run downloads nothing.

### Validation

Cache deleted and re-downloaded from scratch: all 40 files match their original
SHA-256. Same text/seed at `--temperature 0`, vs `main`:

| comparison | result |
| --- | --- |
| multifunction stepped / fused / throughputOptimized vs `main` | bit-exact |
| legacy `W8A16` vs multifunction stepped | corr 0.99999, max abs 7.4e-3 |
| pinned `W8A16` against a cache that already has it | 0 bytes downloaded |

Both guards fire: legacy + `fused` and legacy + `throughputOptimized` each report
the single-function configuration error. 4 tests added for rank-2 geometry. The
macOS 14 path stays deleted — both layouts need the macOS 15 / iOS 18 `MLTensor`
floor.

One commit is unrelated to the decoders: it gates `xcbeautify --renderer
github-actions` to a single matrix job, since eight jobs were annotating the same
warning eight times.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -98,7 +98,8 @@ jobs:
         if: ${{ matrix.run-config['condition'] == true }}
         run: |
           set -o pipefail
-          xcodebuild clean build-for-testing -scheme argmax-oss-swift-Package -destination '${{ matrix.run-config['clean-destination'] }}' | xcbeautify --renderer github-actions
+          # Annotate warnings from a single job only; duplicate check runs each re-post identical annotations on the PR diff
+          xcodebuild clean build-for-testing -scheme argmax-oss-swift-Package -destination '${{ matrix.run-config['clean-destination'] }}' | xcbeautify --renderer ${{ (matrix.run-config['name'] == 'macOS' && inputs.macos-runner == 'macos-26') && 'github-actions' || 'terminal' }}
       - name: Test - ${{ matrix.run-config['name'] }}
         if: ${{ matrix.run-config['condition'] == true }}
         env:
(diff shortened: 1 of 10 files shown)


EduardoPach:
This PR will allow people to do the following in case they have the previous
assets already downloaded and don't want to download the new
`W8A16-multifunction` ones:

```swift
import TTSKit

Task {
    let config = TTSKitConfig(
        multiCodeDecoderVariant: "W8A16",
        speechDecoderVariant: "W8A16"
    )
    let tts = try await TTSKit(config)
    let result = try await tts.generate(text: "Hello from TTSKit!")
    print("Generated \(result.audioDuration)s of audio at \(result.sampleRate)Hz")
}
```

ZachNagengast (approved):

PR #531, opened 2026-09-16
fix(TTSKit): guard against zero probSum in sampleFromProbs to prevent crash

### LINKED ISSUE #478: Fatal crash in GreedyTokenSampler.sampleFromProbs when top-K probability sum is zero

sumitvairagar:
## Summary

Fixes #478.

When all top-K softmax probabilities underflow to zero due to numerical instability (e.g. all logits are `-inf` or extremely negative), `probSum == 0` and `Float.random(in: 0..<0)` triggers a Swift runtime `fatalError`.

## Root Cause

In `sampleFromProbs`, after `probs.topK(topK)` materialises the top-K probabilities, it's possible — under degenerate model outputs — for every value to be `0.0` in Float32. The cumulative-sum sampler then calls:

```swift
let randomValue = Float.random(in: 0..<probSum) // probSum == 0 → fatalError
```

This was introduced when the function was extracted as a shared helper in v1.0.0 and remained present in v1.1.0.

## Fix

Add a `guard probSum > 0` before the random draw. When the guard fires, fall back to **argmax** — the token with the highest probability in the top-K set. This is the correct behaviour: it is the model's most confident choice given that weighted sampling is numerically impossible.

```swift
guard probSum > 0 else {
    let maxIdx = probsArray.indices.max(by: { probsArray[bash] < probsArray[] }) ?? 0
    return Int32(idxArray[maxIdx])
}
```

## Testing

Added `testGreedySamplerZeroProbSumFallsBackToArgmax` to `TTSKitUnitTests`:
- Sets all 16 logits to `-Float.infinity` except token 5 (logit = 1.0)
- Calls `sampleCodec0` with `temperature: 0.9` and `topK: 10`
- Asserts the sampler returns token 5 without crashing
- Guarded with `#available(macOS 15, iOS 18, ...)` to match the MLTensor path


sumitvairagar:
Good catch from Copilot — the original fixture was wrong.

Setting token 5 to `1.0` while the rest were `-infinity` meant softmax gave token 5 a probability of ~1.0, so `probSum` was never zero and the guard never fired.

Fixed in the latest commit: all 16 logits are now `-100.0`. With `temperature=0.9`, `exp(-100/0.9) ≈ 5.6e-49`, which is below Float32 minimum (~1.4e-45) and rounds to exactly `0.0`. Every top-K probability is therefore `0.0`, `probSum == 0`, and the guard fires on every run.

Since all probs are zero the argmax result is arbitrary (first element of top-K), so the assertion now checks `0 <= token < vocabSize` rather than a specific index.

sumitvairagar (commented):
Addressed the Copilot feedback in the second commit (8b5c5bc). The test now sets all 16 logits to `-100.0`, which with `temperature=0.9` produces `exp(-100/0.9) ≈ 5.6e-49` — below Float32 minimum (~1.4e-45), so every top-K probability rounds to exactly `0.0`, `probSum == 0`, and the guard fires on every run. Test confirmed passing locally. Requesting a fresh Copilot review.

sumitvairagar:
Pushed a third commit (b415173) addressing the Copilot feedback.

**The change:** Instead of one sampler asserting a valid range, the test now runs two samplers with different seeds (0 and 99999) on the same all-zero-logit input and asserts they return **identical tokens**.

**Why this proves the guard fired:**
- If `Float.random` was called (RNG path), different seeds would produce different tokens with high probability
- The argmax fallback is fully deterministic — same input always yields the same index regardless of seed
- Equal results from two different seeds = argmax path taken = `guard probSum > 0` fired

The comment also clarifies the Float16→Float32 cast path: the MLMultiArray is `.float16` but `sampleCodec0WithMLTensor` casts to Float32 before softmax, so `exp(-100/0.9) ≈ 5.6e-49 < Float32 min` and softmax rounds to exactly 0.0 as intended.

PR #533, opened 2026-09-16
Stop recording when streaming transcription exits on error

rodrigouroz:
When `AudioStreamTranscriber.realtimeLoop()` catches a transcription error, including cancellation while waiting for audio, it logs the error and exits without stopping capture or resetting `isRecording`. The streaming task has finished, but recording remains active until the caller explicitly stops it, and subsequent starts are blocked by the recording flag.

Call the existing `stopStreamTranscription()` before leaving the loop on error. This stops the audio processor and publishes the recording-state transition. Normal Stop behavior and error propagation remain unchanged; this does not add a final transcription pass or alter segment confirmation.

The recording lifecycle is extracted into an internal method after the system permission check, so regression tests can exercise the production path with an `AudioProcessing` test double, without microphone access or model downloads. No public API is added.

Validation on macOS arm64 with Swift 6.4:

- Before the fix: transcription-error and cancellation tests fail because capture remains active, no stop is requested, and the callback never reports `isRecording == false`. The explicit-Stop test passes.
- After the fix: `swift test --filter AudioStreamTranscriberTests` passes all three tests.
- `make build` and `git diff --check` pass.

The decoder-error test intentionally uses an unloaded decoder to produce a deterministic error. These tests verify lifecycle cleanup; they do not assess speech recognition accuracy or physical audio hardware.

The complete model-dependent test suite and other platforms were not run locally.


ZachNagengast (approved):

ZachNagengast:
Looks good, thanks for the thorough tests!

Issue #491, opened 2026-06-12
Partial model download is never repaired, resolveRepo's local-cache check accepts incomplete folders


ericlewis:
## Summary

If a TTSKit model download is interrupted partway (app backgrounded/killed, network drop), the partially-downloaded folder permanently poisons the cache: every subsequent `setupModels`/`TTSKit.download` call short-circuits on a local-cache check that accepts *any non-empty directory* as complete, so the missing files are never fetched and `loadModels()` fails with Core ML compile errors forever. The only way out is manually deleting the model folder.

## Environment

- argmax-oss-swift **1.0.0** (`25c6299`), product `TTSKit`
- iPhone 17 Pro, iOS 27 beta
- Default config (`.qwen3TTS_0_6b`, repo `argmaxinc/ttskit-coreml`), foreground download session

## Repro

1. Start a fresh model download (`setupModels(download: true)` or first `generate()`).
2. Kill or background the app mid-download (foreground `URLSession`, so transfers stop).
3. Relaunch and load again.

**Expected:** the missing files are detected and re-downloaded (the per-file etag/existence checks in `HubApi.snapshot` handle exactly this).

**Actual:** `resolveRepo` returns the partial folder as "cached", `modelState` proceeds to loading, and `loadModels()` fails:

```
Failed to open file: …/qwen3_tts/code_decoder/12hz-0.6b-customvoice/W8A16-stateful/CodeDecoder.mlmodelc/coremldata.bin. It is not a valid .mlmodelc file.
E5RT: Input model path not found: …/qwen3_tts/text_projector/12hz-0.6b-customvoice/W8A16/TextProjector.mlmodelc/model.mil (1)
Failed to parse ML Program. It is likely an invalid or broken model.
```

On-device contents of `TextProjector.mlmodelc` after the interrupted download — the small files landed, `model.mil` and `weights/` never did:

```
analytics/coremldata.bin    243 B
coremldata.bin              380 B
metadata.json               2 KB
(model.mil missing — 6.5 KB in repo)
(weights/weight.bin missing — 317 MB in repo)
```

## Root cause

[`ModelDownloader.resolveRepo`](https://github.com/argmaxinc/argmax-oss-swift/blob/25c62997041c134b03ca82731ce2f6fd2cae1eb9/Sources/ArgmaxCore/ModelDownloader.swift#L197) short-circuits before `HubApi.snapshot` ever runs:

```swift
if patternsExistLocally(patterns, in: localRoot) {
[...]

PR #534, opened 2026-09-16
Throw when a Hub snapshot download is cancelled

rodrigouroz:
`HubApi.snapshot` can return its repository URL after cancellation even though requested files are still missing. For example, cancelling a task while it waits for the next file ends the download state stream, then the snapshot loop returns the partial folder as success. Replace that return with `Task.checkCancellation()` so the caller receives `CancellationError`.

Refs #491, specifically [the snapshot cancellation report](https://github.com/argmaxinc/argmax-oss-swift/issues/491#issuecomment-4696676741). Offline cache completeness and repair remain separate concerns; this change does not close the entire issue.

The regression test runs the real snapshot and file downloader against a small HTTP server bound to loopback. It starts with one cached file, waits until the missing file request arrives, and cancels while that response is held. It checks the error type and that the cache still contains only the original file. A second test completes the same partial download and verifies both file contents and final progress. Neither test requires Hugging Face access or model assets.

A downstream example is [VoxKey's model installer](https://github.com/rodrigouroz/VoxKey/blob/a5c84a1e692a4cbc8addff5d6e922d5a5c1a4c74/Sources/VoxKeyApp/TranscriptionModelLibrary.swift#L42-L51), which checks cancellation immediately after `WhisperKit.download` and validates required model files before accepting the installation. This change makes the SDK itself report cancellation at this boundary; downstream integrity checks remain useful.

Validation on macOS, Swift 6.4 with the package's Swift 5 language mode:

- Before the fix: the cancellation regression fails with a partial-success result, while the normal download passes. Confirmed in four executions of the final test.
- `swift test --filter 'HubApiTests|ModelDownloaderTests|DownloaderTests'`: 12 tests passed.
- Ten additional runs of `swift test --skip-build --filter HubApiTests`: both tests passed each time.
- `make build` and `git diff --check` passed.

The full model-dependent suite and other Apple platforms were not run locally.


ZachNagengast (approved):

rodrigouroz:
Thanks @ZachNagengast 
I checked the failed test and it seems to be HF rate limiting, I guess we can just retry and it will go green.
Cheers!

